PR #11, opened 2016-02-05
event: Regenerate and fix idiomatic Go naming issues.

### LINKED ISSUE #4: DblClick should be DoubleClick
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L186

### LINKED ISSUE #5: Gotpointercapture should be GotPointerCapture
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L342

### LINKED ISSUE #6: Lostpointercapture should be LostPointerCapture
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L426

### LINKED ISSUE #7: Pointerfoobar should be PointerFooBar
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L562-L608

### LINKED ISSUE #8: ReadystateChange should be ReadyStateChange
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L630

### LINKED ISSUE #9: Selectionchange should be SelectionChange
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L720

### LINKED ISSUE #10: Selectstart should be SelectStart
https://github.com/gopherjs/vecty/blob/master/event/event.gen.go#L726

shurcooL:
Fixes #4.
Fixes #5.
Fixes #6.
Fixes #7.
Fixes #8.
Fixes #9.
Fixes #10.

/cc @slimsag @neelance


slimsag:
The regenerator is broken, this change is invalid. :/ will elaborate in
person tomorrow

event/event.gen.go
```diff
@@ -125,12 +145,6 @@ func Close(listener func(*vecty.Event)) *vecty.EventListener {
 	return &vecty.EventListener{Name: "close", Listener: listener}
 }
 
-// The compass used to obtain orientation data is in need of calibration.
-// https://developer.mozilla.org/docs/Web/Events/compassneedscalibration
-func CompassNeedsCalibration(listener func(*vecty.Event)) *vecty.EventListener {
```
slimsag: why did this dissapear I wonder?

event/event.gen.go
```diff
@@ -545,63 +583,63 @@ func Playing(listener func(*vecty.Event)) *vecty.EventListener {
...
-// https://developer.mozilla.org/docs/Web/Events/pointerlockchange
-func PointerLockChange(listener func(*vecty.Event)) *vecty.EventListener {
-	return &vecty.EventListener{Name: "pointerlockchange", Listener: listener}
-}
-
-// It was impossible to lock the pointer for technical reasons or because the permission was denied.
-// https://developer.mozilla.org/docs/Web/Events/pointerlockerror
-func PointerLockError(listener func(*vecty.Event)) *vecty.EventListener {
-	return &vecty.EventListener{Name: "pointerlockerror", Listener: listener}
-}
```
slimsag: same question for these two?

event/event.gen.go
shurcooL: It looks like it's no longer listed in the table at https://developer.mozilla.org/en-US/docs/Web/Events.

There is still a mention at the sensor events section:

image

But the link is 404:

https://developer.mozilla.org/en-US/docs/Web/Events/compassneedscalibration

event/event.gen.go
```diff
@@ -545,63 +583,63 @@ func Playing(listener func(*vecty.Event)) *vecty.EventListener {
...
+// https://developer.mozilla.org/docs/Web/Events/pointerlockchange
+func PointerLockChange(listener func(*vecty.Event)) *vecty.EventListener {
+	return &vecty.EventListener{Name: "pointerlockchange", Listener: listener}
+}
+
+// It was impossible to lock the pointer for technical reasons or because the permission was denied.
+// https://developer.mozilla.org/docs/Web/Events/pointerlockerror
+func PointerLockError(listener func(*vecty.Event)) *vecty.EventListener {
+	return &vecty.EventListener{Name: "pointerlockerror", Listener: listener}
+}
```
shurcooL: It just got moved here.

slimsag: :+1:

event/event.gen.go
shurcooL: They've moved to [there](https://github.com/gopherjs/vecty/pull/11#discussion_r52086811).

event/event.gen.go
slimsag: I found it! The issue is that it was removed from the standard events table (because Firefox doesn't support it, but IE 11 does). Removing it seems to be correct behavior, so let's merge as-is and reconsider how to handle non-standard events in the future (will open an issue for this)

PR #12, opened 2016-02-08
domutil: merge into top-level package

slimsag:
The `domutil` package is not used by anyone except the top-level `vecty` package. Merge it into the top-level package for now, and consider factoring out into `internal/domutil` later if needed.


slimsag:
@neelance what are your thoughts (you wrote the code so I might be missing a reason for keeping htis as-is)?

neelance:
LGTM. It was used by multiple package in an earlier version.

PR #14, opened 2016-02-08
event: cleanup docs slightly

slimsag:
Cleanup the `event` package docs slightly:

- Less verbosity in linking to MDN.
- Blank line comment for cleaner rendering on godoc.org etc.

### DIFF
--- a/event/generate.go
+++ b/event/generate.go
@@ -163,7 +163,7 @@ func main() {
 
 // Package event defines markup to bind DOM events.
 //
-// See https://developer.mozilla.org/en-US/docs/Web/Events
+// See https://developer.mozilla.org/en-US/docs/Web/Events.
 package event
 
 import (
--- a/event/generate.go
+++ b/event/generate.go
@@ -161,7 +161,9 @@ func main() {
 
 	fmt.Fprint(file, `//go:generate go run generate.go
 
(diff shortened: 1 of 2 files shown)


event/generate.go
```diff
@@ -161,7 +161,9 @@ func main() {
 
 	fmt.Fprint(file, `//go:generate go run generate.go
 
-// Documentation source: "Event reference" by Mozilla Contributors, https://developer.mozilla.org/en-US/docs/Web/Events, licensed under CC-BY-SA 2.5.
+// Package event defines markup to bind DOM events.
+//
+// See https://developer.mozilla.org/en-US/docs/Web/Events
```
shurcooL: Nitpick. Given that this is a sentence, I'd consider not leaving out the period at the end.

neelance: This change will make us violate the MDN license. Let's be nice open source citizens and do proper attribution as described on https://developer.mozilla.org/en-US/docs/MDN/About#Copyrights_and_licenses

slimsag:
Good point @neelance and @shurcooL I hadn't considered the license aspect. I thought it was just a bad documentation line (due to `Documentation source: ...` reading strangely).

I've updated the PR with what I think would be better, PTAL.

neelance:
LGTM

PR #16, opened 2016-02-11
elem: package cleanup

slimsag:


slimsag:
@neelance I found a better solution to the issue with `<image>` `<bgsound>` etc not being marked properly on the page as non-standard etc. I can edit the page because it's a wiki!

I updated all the tags on that page to reflect what they show when you click on them, just waiting for the page to 'rebuild' or something now before I can regenerate. https://developer.mozilla.org/en-US/docs/Web/HTML/Element

image

neelance:
@slimsag Cool!

slimsag:
All done now, PTAL

neelance:
LGTM

PR #17, opened 2016-02-11
do not needlessly export TextComponent

slimsag:


neelance:
LGTM. We should look into reducing the public API even more.

shurcooL:
> We should look into reducing the public API even more.

Huge :+1:.

PR #18, opened 2016-02-13
s/Property/Prop/g

slimsag:
Rename `Property` to just `Prop` to match `vecty/prop` package naming.


slimsag:
Closing in favor of https://github.com/gopherjs/vecty/pull/19

PR #19, opened 2016-02-13
all: minor cleanups

slimsag:


shurcooL:
LGTM. I did not spot any problems that I'd want to comment on.

neelance:
Why the change to `Elem` and `Prop`?

neelance:
Some context: The `Element` and `Property` functions are only rarely used directly, most of the times one uses the `elem` and `prop` packages instead. I also don't have a preference for the longer version. @shurcooL It's up to you, which one do you prefer?

slimsag:
The change to `Elem` and `Prop` was because IMO it makes sense to have them match the packages also named `elem` and `prop`. It's nit-picky, though, and as you said they are often not used directly .. so I'm fine with whatever @shurcooL and you think is best :)

shurcooL:
I'm also fine with either one, it's not clear to me at this time which is better. But I suspect over a longer period of time it may become more clear.

Since we're okay with both options, I'd say let's go with the simplest thing for now. IMO, it's simpler to maintain the longer full name. It avoids having to make the decision to truncate the name at some arbitrary length, which is a decision that should be done when there's a good reason to.

It'd also be inconsistent with `Component` being `Component` and not `Comp`. But `Comp` is confusing.

Let's go with whatever's easiest and simplest for now, and change later if there's a stronger motivation to do so.

shurcooL:
Just to add a point, I don't think that a package name has to match identifier name - I'm completely okay with a difference there. They use different rules. Package names are short, all lower case. Identifiers are MixedCase, often longer and more descriptive names. They're rarely abbreviated. Package names are more commonly abbreviated.

So package name being `elem` and an identifier `Element` is completely consistent and okay by me.

slimsag:
Got it. I'll back out those two `Elem` and `Prop` changes and force-push. Do the other changes LGTY @neelance ?

neelance:
Yeah, rest LGTM.

PR #33, opened 2016-02-21
implement removal of styles

### LINKED ISSUE #26: Fix style reset
See https://github.com/gopherjs/vecty/blob/master/dom.go#L101

@neelance can you shed some light on what this TODO means? I'd like to take a look into it.

slimsag:
Fixes #26

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -101,10 +101,15 @@ func (e *Element) Reconcile(oldComp Component) {
 			}
 		}
 
-		// TODO fix style reset
 		style := e.node.Get("style")
 		for name, value := range e.Style {
 			style.Call("setProperty", name, value)
+		}
+		for name := range oldElement.Style {
+			if _, ok := e.Style[name]; !ok {
+				// Style was removed.
+				style.Call("removeProperty", name)
+			}
 		}
 
 		for _, l := range oldElement.EventListeners {
--- a/dom.go
+++ b/dom.go
@@ -107,7 +107,6 @@ func (e *Element) Reconcile(oldComp Component) {
 		}
 		for name := range oldElement.Style {
 			if _, ok := e.Style[name]; !ok {
-				// Style was removed.
 				style.Call("removeProperty", name)
 			}
 		}


slimsag:
This is basically untested .. I don't think I can trigger this with the TodoMVC app, right?

Either way, I'm pretty sure this is correct (and it's better than a TODO). See [the poor MDN docs here](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration) or a slightly [better / weird docpage here](http://help.dottoro.com/ljopsjck.php). The only downside to this approach is a strict requirement on IE 9+, but I don't think anyone (myself included) will object to this ;)

dom.go
```diff
@@ -101,11 +101,16 @@ func (e *Element) Reconcile(oldComp Component) {
...
 		}
 
-		// TODO fix style reset
 		style := e.node.Get("style")
 		for name, value := range e.Style {
 			style.Call("setProperty", name, value)
 		}
+		for name := range oldElement.Style {
+			if _, ok := e.Style[name]; !ok {
+				// Style was removed.
```
shurcooL: This comment seems unhelpful, I'd suggest removing it.

If you look at the greater context of this func, other similar sections (for properties, event listeners, etc.) do not have an equivalent comment. I don't think they need it either, at least not at this level. The actual single line of code below is equally descriptive IMO.

shurcooL:
> This is basically untested

Perhaps it's useful to add a "// TODO: Verify this works, then remove this comment." comment then? That way, if someone runs into an issue here, they'll know to report it immediately instead of wasting time thinking they're doing something wrong.

The logic seems sound to me, so LGTM.

dom.go
slimsag: I'll remove it for now and reconsider docs within this func later.

slimsag:
> Perhaps it's useful to add a "// TODO: Verify this works, then remove this comment." comment then? That way, if someone runs into an issue here, they'll know to report it immediately instead of wasting time thinking they're doing something wrong.

> Edit: To force it to be more actionable and help the above TODO get resolved faster, perhaps it can be an log.Println("TODO: Verify this works as expected, then remove this log statement.") instead. People who run into the message will be motivated to resolve the TODO since it'll litter their stdout. :) This is inspired by goxjs/gl#15 (comment).

Honestly.. this is very off-putting to me. I don't like TODO's in code unless they are purely for markers (if real actionable items) and littering stdout doesn't seem very nice to me either. I'm coming from the stance that this code is _most likely_ correct and a TODO or log line is probably not worth the trouble of just removing it later. If this code doesn't work, anyone can easily open an issue and we'll know right away just as easily (if you remove a style, it should be removed from the browser DOM, just as you'd expect).

I personally feel comfortable committing the code as-is right now (i.e. find it unlikely that it would be incorrect), but if we don't want to go based on feelings I'd rather not merge the change at all (until it can be tested by someone, I mean, rather than littering the code).

shurcooL:
> I'm coming from the stance that this code is most likely correct and a TODO or log line is probably not worth the trouble of just removing it later.
> ...
> I personally feel comfortable committing the code as-is right now (i.e. find it unlikely that it would be incorrect)

In that case, please don't use my suggestion, since it's not helpful. My LGTM still stands.

neelance:
LGTM

slimsag:
(just got around to checking back on this)

> In that case, please don't use my suggestion (I used the word "perhaps" to indicate it's optional), since it's not beneficial. My LGTM still stands.

@shurcooL hope my comment came across as more of an explanation/justification for me wanting to do it a different way, rather than coming across as rude / or uninterested. If it was the latter, I apologize :smile:

shurcooL:
> hope my comment came across as more of an explanation/justification for me wanting to do it a different way

Absolutely. It was a good argument, and it convinced me.

When I said:

> In that case

What I meant by that was:

> I agree with your arguments, and now believe that your suggestion is better, so disregard my previous suggestion since I now think it's actually worse.

> coming across as rude / or uninterested

Not at all! :+1: We're both striving to find the best solution, and I think we came to a consensus on what that was. :)

shurcooL:
Also, I'm glad that my suggestion of adding a TODO led you @slimsag to think it over and demonstrate confidence in the current code not needing a TODO, which led to better final code. Originally, you weren't as confident about it. :P

PR #34, opened 2016-02-21
added some more styles

neelance:


style/style.go
```diff
@@ -16,14 +16,52 @@ func Color(value string) vecty.Markup {
...
+	return vecty.Style("overflow", option)
+}
+
+func OverflowX(option OverflowOption) vecty.Markup {
+	return vecty.Style("overflow-x", option)
+}
+
+func OverflowY(option OverflowOption) vecty.Markup {
+	return vecty.Style("overflow-y", option)
+>>>>>>> Stashed changes
```
shurcooL: ">>>>>>> Stashed changes" doesn't belong here.

style/style.go
```diff
@@ -16,14 +16,52 @@ func Color(value string) vecty.Markup {
...
 
-func Width(size Size) vecty.Markup {
-	return vecty.Style("width", string(size))
+type OverflowOption string
+
+const (
+	OverflowVisible OverflowOption = "visible"
+	OverflowHidden                 = "hidden"
+	OverflowScroll                 = "scroll"
+	OverflowAuto                   = "auto"
```
shurcooL: I believe your intent was to make all these `const`s have type `OverflowOption`.

However, since you're not using `iota`, only the first entry has type `OverflowOption`. All others are untyped strings. Fix it by doing:

```Go
const (
	OverflowVisible OverflowOption = "visible"
	OverflowHidden  OverflowOption = "hidden"
	OverflowScroll  OverflowOption = "scroll"
	OverflowAuto    OverflowOption = "auto"
)
```

An alternative form is:

```Go
const (
	OverflowVisible = OverflowOption("visible")
	OverflowHidden  = OverflowOption("hidden")
	OverflowScroll  = OverflowOption("scroll")
	OverflowAuto    = OverflowOption("auto")
)
```

But I don't think it's necessarily better. Just sharing for posterity.

shurcooL:
2 issues to fix (see inline comments above), then LGTM.

shurcooL:
LGTM.

PR #35, opened 2016-02-21
added support for custom data attributes

neelance:


markup.go
```diff
@@ -37,6 +37,28 @@ func Property(name string, value interface{}) Markup {
...
+}
+
+// Apply implements the Markup interface.
+func (d *data) Apply(element *Element) {
+	if element.Dataset == nil {
+		element.Dataset = make(map[string]string)
+	}
+	if _, ok := element.Dataset[d.name]; ok {
+		panic(fmt.Sprintf("duplicate data: %s", d.name))
+	}
```
shurcooL: This is very minor. But I want to point out that it's possible to perform this duplicate data check **before** the `if element.Dataset == nil { element.Dataset = make(map[string]string) }` block above.

shurcooL:
LGTM. Read 1 comment above.

PR #39, opened 2016-04-23
examples/todomvc: properly focus the text input when double clicking on an item

### LINKED ISSUE #21: browser DOM refs
The TodoMVC example currently does not select the text input when double clicking on an item. This is because we don't have any way to call `theInputElement.focus()` which in React would look something like:

```
ReactDOM.findDOMNode(this.refs.theInputElement).focus(); 
```

Need to figure out the right design for this.

slimsag:
Closes #21

### DIFF
--- a/examples/todomvc/components/itemview.go
+++ b/examples/todomvc/components/itemview.go
@@ -18,6 +18,7 @@ type ItemView struct {
 	Item      *model.Item
 	editing   bool
 	editTitle string
+	input     *vecty.Element
 }
 
 // Apply implements the vecty.Markup interface.
@@ -52,6 +53,7 @@ func (p *ItemView) onStartEdit(event *vecty.Event) {
 	p.editing = true
 	p.editTitle = p.Item.Title
 	p.ReconcileBody()
+	p.input.Node().Call("focus")
 }
 
 func (p *ItemView) onEditInput(event *vecty.Event) {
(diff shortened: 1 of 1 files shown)


shurcooL:
I'm not closely familiar with design principles and whether this is the correct approach to focusing elements in a react-style package.

But the change is logical and makes sense to me, and I'd expect it to work. You keep a reference to the input element and call focus on it on start edit.

LGTM, if you're confident about the approach.

examples/todomvc/components/itemview.go
```diff
@@ -52,6 +53,7 @@ func (p *ItemView) onStartEdit(event *vecty.Event) {
 	p.editing = true
 	p.editTitle = p.Item.Title
 	p.ReconcileBody()
+	p.input.Node().Call("focus")
```
shurcooL: This is guaranteed to happen after `ItemView.render()` was called, right? Otherwise `p.input` will be `nil`.

slimsag:
Totally not confident in the approach yet; I mean it works, but there could be caveats I haven't foreseen. Maybe `this.refs` in React does something more complex than this. It'd be good if someone with React experience chimed in, otherwise I'll have to research myself at a later time.

examples/todomvc/components/itemview.go
slimsag: yeah, because `onStartEdit` is only called by an event handler which is itself registered on double click within the `render` method itself.

slimsag:
Perhaps @neelance knows?

neelance:
I'm also not 100% sure, but this solution seems simple and functional, so I would opt for keeping it until we know better. Vecty is inspired by React, but it already has some significant differences. React needs refs because the data structures that the render function returns are really just placeholders for the actual elements. Vecty's approach is simpler and makes this solution possible. Nice, @slimsag!

PR #40, opened 2016-04-23
component architecture changes

slimsag:
Over the past months I've been looking into the component architecture, and in particular, how to document it and explain it to others efficiently. I've tried probably 7+ different approaches at this point, and this is the one that I've settled into / feel the most comfortable with.

Start by reading the commit message of f91790a to get an idea for motivation, and please compare the todomvc example after that change to the todomvc example before that change.

I would be very curious to get the opinions of @shurcooL and @neelance here, although I understand this is a more controversial change and probably much less likely to be accepted. I'm okay with that and with this change taking longer to review, potentially not being merged, etc.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -2,6 +2,48 @@ package vecty
 
 import "github.com/gopherjs/gopherjs/js"
 
+// Renderable is a user-defined renderable component.
+type Renderable interface {
+	Component
+	Render() Component
+}
+
+// Core is the struct which all components embed.
+type Core struct {
+	component Renderable
+	body      Component
+}
+
+// Apply implements the Markup interface.
+func (c *Core) Apply(e *Element) {
+	e.AddChild(c.component)
+}
+
+// Node implements the Component interface.
+func (c *Core) Node() *js.Object {
+	return c.body.Node()
+}
+
+// Reconcile implements the Component interface.
+func (c *Core) Reconcile(oldComp Component) {
+	oldBody := c.body
+	c.body = c.component.Render()
+	c.body.Reconcile(oldBody)
+	if oldBody != nil {
+		replaceNode(c.body.Node(), oldBody.Node())
+	}
+}
+
+// Rerender causes the component to rerender the body and reconcile it.
+func (c *Core) Rerender() {
+	c.Reconcile(nil)
+}
+
+// New creates a new Core component object.
+func New(component Renderable) *Core {
+	return &Core{component: component}
+}
+
 // Component represents a Vecty component.
 type Component interface {
 	Markup
@@ -172,25 +214,4 @@ func AddStylesheet(url string) {
 	link.Set("rel", "stylesheet")
 	link.Set("href", url)
 	js.Global.Get("document").Get("head").Call("appendChild", link)
-}
-
-// Composite is the struct which all components embed.
-type Composite struct {
-	RenderFunc func() Component
-	Body       Component
-}
-
-// Node implements the Component interface.
-func (c *Composite) Node() *js.Object {
-	return c.Body.Node()
-}
-
-// ReconcileBody implements the Component interface.
-func (c *Composite) ReconcileBody() {
-	oldBody := c.Body
-	c.Body = c.RenderFunc()
-	c.Body.Reconcile(oldBody)
-	if oldBody != nil {
-		replaceNode(c.Body.Node(), oldBody.Node())
-	}
 }
--- a/dom.go
+++ b/dom.go
@@ -19,6 +19,15 @@ func (c *Core) Apply(e *Element) {
 	e.AddChild(c.component)
 }
 
+// Unmount implements the Component interface.
+func (c *Core) Unmount() {
+	// Break the circular link between c (Core) and c.component (Renderable) or
+	// else one object could not be garbage collector without the other (both
+	// would have to be unused).
+	c.component = nil
+	c.body.Unmount()
+}
+
 // Node implements the Component interface.
 func (c *Core) Node() *js.Object {
 	return c.body.Node()
@@ -47,6 +56,11 @@ func New(component Renderable) *Core {
 // Component represents a Vecty component.
 type Component interface {
 	Markup
+
+	// Unmount is called when the component should be unmounted. i.e., when all
+	// of its event listeners and DOM elements should be removed.
+	Unmount()
+
 	Reconcile(oldComp Component)
 	Node() *js.Object
 }
@@ -74,6 +88,11 @@ type textComponent struct {
 // Apply implements the Markup interface.
 func (s *textComponent) Apply(element *Element) {
 	element.Children = append(element.Children, s)
+}
+
+// Unmount unmounts this textComponent by removing its node from the DOM.
+func (s *textComponent) Unmount() {
+	removeNode(s.node)
 }
 
 func (s *textComponent) Reconcile(oldComp Component) {
@@ -117,6 +136,17 @@ func (e *Element) AddChild(s Component) {
 // Apply implements the Markup interface.
 func (e *Element) Apply(element *Element) {
 	element.Children = append(element.Children, e)
+}
+
+// Unmount unmounts this Element component by removing its node from the DOM.
+func (e *Element) Unmount() {
+	for _, child := range e.Children {
+		child.Unmount()
+	}
+	for _, l := range e.EventListeners {
+		e.node.Call("removeEventListener", l.Name, l.wrapper)
+	}
+	removeNode(e.node)
 }
 
 // Reconcile implements the Component interface.
@@ -173,7 +203,7 @@ func (e *Element) Reconcile(oldComp Component) {
 			replaceNode(newChild.Node(), oldChild.Node())
 		}
 		for i := len(e.Children); i < len(oldElement.Children); i++ {
-			removeNode(oldElement.Children[i].Node())
+			oldElement.Children[i].Unmount()
 		}
 		return
 	}
--- a/dom.go
+++ b/dom.go
@@ -57,8 +57,8 @@ func New(component Renderable) *Core {
 type Component interface {
 	Markup
 
-	// Unmount is called when the component should be unmounted. i.e., when all
-	// of its event listeners and DOM elements should be removed.
+	// Unmount is called before the component is unmounted. i.e., when all of
+	// its event listeners and DOM elements will be removed.
 	Unmount()
 
 	Reconcile(oldComp Component)
@@ -90,10 +90,8 @@ func (s *textComponent) Apply(element *Element) {
 	element.Children = append(element.Children, s)
 }
 
-// Unmount unmounts this textComponent by removing its node from the DOM.
-func (s *textComponent) Unmount() {
-	removeNode(s.node)
-}
+// Unmount implements the Component interface and is no-op.
+func (s *textComponent) Unmount() {}
 
 func (s *textComponent) Reconcile(oldComp Component) {
 	if oldText, ok := oldComp.(*textComponent); ok {
@@ -138,7 +136,8 @@ func (e *Element) Apply(element *Element) {
 	element.Children = append(element.Children, e)
 }
 
-// Unmount unmounts this Element component by removing its node from the DOM.
+// Unmount unmounts this Element component by calling unmount on all children
+// and removing all event listeners.
 func (e *Element) Unmount() {
 	for _, child := range e.Children {
 		child.Unmount()
@@ -146,7 +145,6 @@ func (e *Element) Unmount() {
 	for _, l := range e.EventListeners {
 		e.node.Call("removeEventListener", l.Name, l.wrapper)
 	}
-	removeNode(e.node)
 }
 
 // Reconcile implements the Component interface.
@@ -204,6 +202,7 @@ func (e *Element) Reconcile(oldComp Component) {
 		}
 		for i := len(e.Children); i < len(oldElement.Children); i++ {
 			oldElement.Children[i].Unmount()
+			removeNode(oldElement.Children[i].Node())
 		}
 		return
 	}
--- a/dom.go
+++ b/dom.go
@@ -1,6 +1,10 @@
 package vecty
 
-import "github.com/gopherjs/gopherjs/js"
+import (
+	"reflect"
+
+	"github.com/gopherjs/gopherjs/js"
+)
 
 // Renderable is a user-defined renderable component.
 type Renderable interface {
@@ -30,18 +34,29 @@ func (c *Core) Node() *js.Object {
 }
 
 // Reconcile implements the Component interface.
-func (c *Core) Reconcile(oldComp Component) {
+func (c *Core) Reconcile(prev Component) {
+	// This function is the generic equivalent to:
+	//
+	// 	if p, ok := prev.(*UserComponent); ok {
+	// 		c.body = oldComp.body
+	// 	}
+	// 	b.Rerender()
+	//
+	if reflect.TypeOf(c.component) == reflect.TypeOf(prev) {
+		prevCore := reflect.Indirect(reflect.ValueOf(prev)).FieldByName("Core").Interface().(*Core)
+		c.body = prevCore.body
+	}
+	c.Rerender()
+}
+
+// Rerender causes the component to rerender the body and reconcile it.
+func (c *Core) Rerender() {
 	oldBody := c.body
 	c.body = c.component.Render()
 	c.body.Reconcile(oldBody)
 	if oldBody != nil {
 		replaceNode(c.body.Node(), oldBody.Node())
 	}
-}
-
-// Rerender causes the component to rerender the body and reconcile it.
-func (c *Core) Rerender() {
-	c.Reconcile(nil)
 }
 
 // New creates a new Core component object.
--- a/dom.go
+++ b/dom.go
@@ -38,7 +38,7 @@ func (c *Core) Reconcile(prev Component) {
 	// This function is the generic equivalent to:
 	//
 	// 	if p, ok := prev.(*UserComponent); ok {
-	// 		c.body = oldComp.body
+	// 		c.body = p.body
 	// 	}
 	// 	b.Rerender()
 	//
--- a/dom.go
+++ b/dom.go
@@ -21,10 +21,6 @@ func (c *Core) Apply(e *Element) {
 
 // Unmount implements the Component interface.
 func (c *Core) Unmount() {
-	// Break the circular link between c (Core) and c.component (Renderable) or
-	// else one object could not be garbage collector without the other (both
-	// would have to be unused).
-	c.component = nil
 	c.body.Unmount()
 }
 
--- a/examples/todomvc/components/pageview.go
+++ b/examples/todomvc/components/pageview.go
@@ -23,7 +23,7 @@ type PageView struct {
 
 func (p *PageView) Reconcile(prev vecty.Component) {
 	if v, ok := prev.(*PageView); ok {
-		v.newItemTitle = v.newItemTitle
+		p.newItemTitle = v.newItemTitle
 	}
 	p.Core.Reconcile(prev)
 }
--- a/examples/todomvc/components/pageview.go
+++ b/examples/todomvc/components/pageview.go
@@ -15,29 +15,22 @@ import (
 )
 
 type PageView struct {
-	vecty.Composite
+	*vecty.Core
 
 	Items        []*model.Item
 	newItemTitle string
 }
 
-// Apply implements the vecty.Markup interface.
-func (p *PageView) Apply(element *vecty.Element) {
-	element.AddChild(p)
-}
-
-func (p *PageView) Reconcile(oldComp vecty.Component) {
-	if oldComp, ok := oldComp.(*PageView); ok {
-		p.Body = oldComp.Body
-		p.newItemTitle = oldComp.newItemTitle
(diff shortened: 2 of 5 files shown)


shurcooL:
I have thought about and prototyped some prerequisite components on which a React-like library would be built, but I don't have much concrete recent experience trying to apply it to React-like architectures. So take my comments FWIW. My thoughts should hold much less weight than those of someone with a closer familiarity, which includes you Stephen since you've actually "tried probably 7+ different approaches". :)

> If instead we define a `Core` struct with basic or 'default' implementations of
> these methods (`Apply`, `Reconcile` and the `RenderFunc`) then we can put
> ourselves in a position where we describe these methods as an opt-in 'feature'
> instead of an up-front learning cost. This makes it easier to learn Vecty right
> off the bat.
> 
> The only caveat is that it's hard to write a `Render` function at an embedded
> `*vecty.Core` struct level, because there is no access to the parent struct,
> the typical problem of trying to map a `super` OOP model to a non-OOP language
> like Go. To address this, we do require that users write some boilerplate
> around component creation by writing their own `NewMyComponent` function:
> 
> ```
> func NewMyComponent(props *MyComponentProps) *MyComponent {
> 	c := &MyComponent{
> 		MyComponentProps: props,
> 	}
> 	c.Core = vecty.New(c)
> 	return c
> }
> ```
> 
> This way the `*vecty.Core` struct has a reference to the parent struct / 'super class'
> if you will.

This makes a lot of sense to me, and I think it's a big win. :+1:

Does this allow getting rid of any of the generated code, or is there no change there?

slimsag:
> Does this allow getting rid of any of the generated code, or is there no change there?

Vecty already has no generated code.. ;) Richard solved that. So no change there, hehe.

shurcooL:
Ah, that's excellent!

bep:
I'm plundering with some GopherJS bindings for Facebook React. It may or may not happen (I'm no React expert), but I have some fun and some frustration doing it. 

But in that process, I follow Vecty with interests. 

The Facebook component lifecycle API is much more complex than Vecty's (kudos to Vecty for that), and I'm still debating with myself about how much of it do I really want to expose.

But I have concluded (before I change my mind again tomorrow) that the cleanest way of doing this is by interface upgrades in Go.

So the most basic component just implements the `Renderer` interface (or whatever).

So quickly translation of my thoughts into Vecty lingo:

```go
func NewMyComponent(props Props,  renderer Renderer) *MyComponent {
    switch i.(renderer) {
        case ComponentWillMount:
           // ... 
        case ComponentDidMount:
           // ...
    }
    c := &MyComponent{
      props: props,
      ...
    }
}
```

slimsag:
@bep interesting thoughts, glad to hear you are looking into something like this too. I think it's long overdue.

But to be somewhat frank here, your code snippet mostly seems to discuss how the different component lifecycle methods are accessed / implemented. In contrast, this PR is _mostly_ about how some component boilerplate like `Apply` and `RenderFunc` are set up. It does include a sort of 'method overriding' scheme, too, so I assume that is what your comment was directed towards (in specific, how you can override `Reconcile`, `Unmount`, and other methods).

I'm curious, within your switch statement what would happen? Would you register a callback? something else?

shurcooL:
> So quickly translation of my thoughts into Vecty lingo:

@bep, sorry, but the Go code you posted does not compile (in my head), meaning I can't understand what exactly you meant. You wrote `switch i.(renderer)` but there is no `i`. Did you mean to write `switch renderer.(type)`?

bep:
>renderer.(type)?

Yes... Sorry about that, typing code directly into a GitHub form isn't my talent. 

>I'm curious, within your switch statement what would happen? Would you register a callback? something else?

In my case I interact with an existing library, so for the relavent lifecycle methods, I do a `js.MakeFunc` which delegates to the user provided implementation. 

I agree it might be side-tracking this topic a little, just felt I should chime in.

slimsag:
Got it, that makes sense. I appreciate you chiming in / giving some feedback.

dom.go
```diff
@@ -173,7 +203,7 @@ func (e *Element) Reconcile(oldComp Component) {
 			replaceNode(newChild.Node(), oldChild.Node())
 		}
 		for i := len(e.Children); i < len(oldElement.Children); i++ {
-			removeNode(oldElement.Children[i].Node())
+			oldElement.Children[i].Unmount()
```
neelance: In this function there is `appendChild` and `replaceNode`, where it is the parent's responsibility to manage the proper insertion of the child's DOM node into the document. It is very inconsistent to not have `removeNode` here as well. We can still have some method to inform the child that it is about to be unmounted, but the actual `removeNode` should not be done by the child.

examples/todomvc/components/pageview.go
```diff
@@ -15,37 +15,30 @@ import (
...
-	if oldComp, ok := oldComp.(*PageView); ok {
-		p.Body = oldComp.Body
-		p.newItemTitle = oldComp.newItemTitle
+func (p *PageView) Reconcile(prev vecty.Component) {
+	if v, ok := prev.(*PageView); ok {
+		v.newItemTitle = v.newItemTitle
 	}
-	p.RenderFunc = p.render
-	p.ReconcileBody()
+	p.Core.Reconcile(prev)
```
neelance: We're loosing explicitness here. For example it is not possible any more to easily include logic to not call `ReconcileBody` if a re-rendering is deemed unnecessary. Adding this feature would require additional API instead of a simple `if` statement previously.

neelance:
I mainly looked at the changes to the todomvc example, because the external API is the important part. I see what you were trying to achieve and it looks quite similar to an earlier state that I had with `neelance/dom`. However, after some time I realized that this is adding "magic" to reduce some lines of boilerplate. You're loosing some useful explicitness (see in-line comment) and lines like `b.Core = vecty.New(b)` initially seem clever (had the same earlier), but make it harder for the user of our library to reason about the control flow. We'll probably have to wait for generics until this boilerplate can be removed in a non-magic way.

I'm still open to be convinced, but for that please answer these questions: What's the advantage to the user who does not know anything about the internals of vecty? Is it easier or harder for him to understand what is going on? Do we gain or loose simplicity?

PS: I really want to avoid the "AngularJS situation" where as soon as the project becomes more complex, you still know which knobs to turn until it behaves as desired after some trial and error, but you don't *really* know what's going on any more. I was there once, didn't like it. React was already much better in that regard.

dom.go
slimsag: Good point.

examples/todomvc/components/pageview.go
slimsag: For my personal clarity / understanding, could you post a snippet of what that logic looked like before this change?

I may be wildly mistaken, but I think it would simply look something like this:

```
if renderIsNeeded {
    p.Core.Reconcile(prev)
}
```

(btw, unrelated and just a friendly note, but you are writing `loose` and `loosing`, like "these pants are too loose".. I think you mean `lose` like "we are losing the race!")

slimsag:
> We'll probably have to wait for generics until this boilerplate can be removed in a non-magic way.

I agree.

> I'm still open to be convinced, but for that please answer these questions: What's the advantage to the user who does not know anything about the internals of vecty? Is it easier or harder for him to understand what is going on? Do we gain or loose simplicity?

These are great questions, and I will get back to them once I figure out the best way to articulate my response here. Possibly next weekend.

> PS: I really want to avoid the "AngularJS situation" where as soon as the project becomes more complex, you still know which knobs to turn until it behaves as desired after some trial and error, but you don't really know what's going on any more. I was there once, didn't like it. React was already much better in that regard.

I should note I don't have much familiarity with the AngularJS situation, but I've heard it mentioned a lot. I highly agree about the understanding aspect, though, and I want to avoid that at all costs too.

davelondon:
Hey guys, I've been playing with vecty over the last couple of months, and quickly found I needed a Unmount hook for getting non-trivial flux notifications working. I've never used react, and my understanding of the vecty internals isn't 100%, so what I've written is doubtless flawed. However, you might be interested in seeing [what I came up with](https://github.com/davelondon/vecty/commit/947b4bc9732277b3cd2d5763f590e81254b9cc62) - I tried to keep it as simple as possible. You can see it in action in my [flux package](https://github.com/kego/ke/tree/vecty/flux), and here's the [beginnings of an app using it](https://github.com/kego/ke/tree/vecty/editor/client).

I'm eager to delete my fork as soon your eventual solution is merged!

slimsag:
@neelance sorry for the lengthy reply, but I wanted to make sure I articulate why I feel very strongly that this is a good change. I'm glad to be wrong, too, as long as you do see my perspective here fully. Most importantly I care that we have the right architecture here, so please take the time to fully read this message whenever you have time (I'm in no rush).

> What's the advantage to the user who does not know anything about the internals of vecty?

- The primary advantage is from both a _documentation_ and _learning curve_ perspective.
- I have considered writing documentation to describe how new users would learn Vecty from ground zero both _with and without this change_, but it's too time-consuming for me to do right now (though, I might do this later, if it would help you see my perspective).
- Consider this outline of some documentation for each perspective (with and without this change):


#### Without this change: Writing your first Component

1. You need a struct, e.g. `PageView`, which embeds `vecty.Composite`.
2. You need an `Apply` method (user thinking: "okay, what does it do?")
  - You can just copy+paste the `AddChild` implementation.
  - In most (all?) cases, you won't ever need to modify the `Apply` method. It's just boilerplate.
  - If you want to understand why Components need an `Apply` method, even though it only does one thing (append your component to an element), then check out this other link which describes how Vecty internally handles things like applying CSS styles to elements.
  - (user thinking: "okay, so just copy+paste it.. got it.")
3. You need an `Reconcile` method
  - This is how Vecty is able to apply the difference between two renders of your component to the browser's DOM (they understand React at a basic, level, so this explanation makes sense to the user).
  - Within this code, you would copy your component's state over: we'll cover this at a later time when your component actually has state.
  - You also need to set your `RenderFunc` and call `ReconcileBody` here. (note: whatever explanation we use for why these are needed is basically a rehash of "this is how Vecty is able to apply the difference between two renders of your component").
  - (user thinking: "okay, so just copy+paste it.. for now.. got it.")
4. You need a `render` method.
  - This is where all the cool stuff happens, now you get to write some HTML code that your component will render as.


#### With this change: Writing your first Component

1. You need a struct, e.g. `PageView`, which embeds `*vecty.Core`.
2. You need a 'constructor' method (see below[4]) (whether we do or don't explain why, the user still thinks "okay, so just copy+paste it.. got it."):
3. You need a `render` method.
  - This is where all the cool stuff happens, now you get to write some HTML code that your component will render as.

[4] the 'constructor' method:
```
func NewPageView() *PageView {
	p := &PageView{}
	p.Core = vecty.New(p)
	return p
}
```

The point is: you get to the real value of Vecty and get a better understanding of the topic at hand because we're only describing things that _you need to know right away_.


> Is it easier or harder for him to understand what is going on?

_Initially_ it is easier for him to understand what is going on, because we're not dancing around topics like "Apply does this.. Reconcile does this.. these are important _but not important to you right now._"

_Long term_ it makes no difference, you still need to understand how `Apply` and `Reconcile` work to do certain things. _But_ we get to explain those _when he needs them_ rather than as an up-front introductory cost.


> Do we gain or loose simplicity?

I would argue that:

1. We gain simplicity by enabling ourselves to explain the parts important to the user up-front, and letting them 'opt-in' to the other methods _on demand / when they need them_, rather than forcing the user to learn these right off the bat.
2. We gain simplicity because there is less duplicated code (the default implementation of these `Apply` and `Reconcile` methods by `Core` reduces the need to copy+paste _some_ code).
3. Without this change, we have to explain why you need `vecty.Composite`, what `ReconcileBody` does and why you must set a `RenderFunc`. With this change, we must explain why you need a `NewPageView` function and why you must initialize `Core` in a convoluted way. These are the same in terms of complexity.

bep:
I still argue that you are doing it harder than it should be. From an end user and a documentation perspective, I would much prefer this:

## Writing your first Component

1. You need a struct, e.g. PageView.
2. You need to at least implement the `Renderer` interface.

All this `Core` emdbedding etc. is something that can happen behind the scenes.

So:

```go

type PageView int

func (p PageView) Render(...) {
// render logic
}

// Build checks what interfaces are implemented, adds sane defaults for the rest.
myComponent := vecty.Build(new(PageView))
```

There may be details lost in the above picture, but it *is* easier to understand and document.

examples/todomvc/components/pageview.go
neelance: Yeah, something like this.

> (btw, unrelated and just a friendly note, but you are writing loose and loosing, like "these pants are too loose".. I think you mean lose like "we are losing the race!")

Thanks a lot for making me aware! This was clearly wrong in my head.

neelance:
@slimsag You're mostly saying that the initial way to start with Vecty is "copy+paste and modify" and I fully agree with that, since this is also how I explore new libraries. I would say that for this step it does not make that much difference if I have to copy 5 lines or 10 lines. But as a next step, I personally want to deeply understand how things fit together, still without having to read the actual implementation. I want to be able to get a good mental model of the internals by looking at the API and how I interact with it.

I still have a not so good gut feeling about this change, but let's first iterate on the implementation and then I'll evaluate it again. I'll post some in-line comments.

dom.go
```diff
@@ -2,9 +2,65 @@ package vecty
...
+}
+
+// Apply implements the Markup interface.
+func (c *Core) Apply(e *Element) {
+	e.AddChild(c.component)
+}
+
+// Unmount implements the Component interface.
+func (c *Core) Unmount() {
+	// Break the circular link between c (Core) and c.component (Renderable) or
```
neelance: This is not true. Garbage collection can recognize circles and remove both. We would be doomed if that wouldn't be the case. ;-)

dom.go
```diff
@@ -2,9 +2,65 @@ package vecty
...
+	c.body.Unmount()
+}
+
+// Node implements the Component interface.
+func (c *Core) Node() *js.Object {
+	return c.body.Node()
+}
+
+// Reconcile implements the Component interface.
+func (c *Core) Reconcile(oldComp Component) {
```
neelance: Why is there `oldComp Component` here and not used? This looks like the old `ReconcileBody` function, but now it is mixed with the "reconcile component" part. This looks strange to me, since those are two distinct steps:
1. You ask a sub-component to reconcile with the previous instance of the sub-component.
2. The component does that and then reconciles its body by rendering and then asking the body to reconcile with the old body (step 1 again).

influx6:
Sorry for being late to the party, being abit coupled down with moving hence my inactivity. Loved the new approach and ideas but I agree with @bep on this one, why working on Haiku, I realised the complexity of that approach and trying to componentize the case through construction functions.

After I killed Haiku and rethought the logic, the interface approach makes sense, it gives users the ability to define the lifecycle of their components without really touching the simplicity of how the Vecty or Haiku view layer worked.

Let your users define their lifecycle and have vecty internally call those to define them, after that separation.

For example: https://github.com/influx6/pocket/blob/master/api/budgets/budget.go

```// Render returns the markup defined for a budget item which is to be rendered.
func (b *Budget) Render() gutrees.Markup {
	root := elems.Div(attrs.Class("budget"))

	barView := elems.Div(attrs.Class("budget-bar", "side-left"))
	barItems := elems.Div(attrs.Class("budget-items", "side-right"))

	for _, item := range b.items {
		item.Render().Apply(barItems)
	}

	barView.Apply(root)
	barItems.Apply(root)

	return root
}```

By having your users define their own behaviour, you do provide a separation of concern and allow them to rebuild new components off those. It simplifies alot.

PS: I have heavily been busy so all these is highly WIP right now, and am also moving houses so its like a rail train hits me everyday. Sorry I have not been as active as I would like but hope this helps in reaching a consesus about what the new vecty feel should be like.

**Comment by influx6** [comment]
Composite components can take the approach to embedded them or provided method overrides, these would allow that simplicity without much complexity for the user.

File: dom.go (hunk 1 above)
Comment (slimsag) [inline dom.go]: Fixed in 5d56a90127582094c8f0d28fb633b85a24b06576

File: dom.go (hunk 3 above)
Comment (slimsag) [inline dom.go]: Yeah, I got confused here. Fixed in e0edc34d2bb62f590b35a6639bb967c2a546d256

File: dom.go (hunk 4 above)
Comment (slimsag) [inline dom.go]: You're 100% correct; and I think this may be the downfall of this PR. If I had to guess, you'll be strongly opposed to my last commit b947f2039e5fa771e3c4d56e2c248790f34ebb02 which fixes this. The alternative to this commit would be requiring users define the `Reconcile` boilerplate themselves.. but at that point the rest of this PR isn't super valuable (aside from Unmount, of course).

I'm also inherently opposed to this approach... just because `reflect` seems cheaty / nasty.. but I'd be interested to hear unbiased opinions? @shurcooL can give thoughts too maybe?

**Comment by slimsag** [comment]
Sorry it's taken so long to update this, I've been quite distracted outside OSS work as of late. @neelance please take another look.

@influx6 @bep I think both of your interface-level suggestions have left out two critical components of the architecture that are mandatory in any complete implementation:

1. How data flows: i.e. how React properties flow from parent components into child components, and how data is accessed externally e.g. like at https://github.com/gopherjs/vecty/blob/master/examples/todomvc/example.go#L22-L27
2. How a component itself requests that its body be reconciled / rerendered upon some state change (e.g. a click event), which prior to this PR is done via an embedded `Composite` struct and in this PR is done via an embedded `Core` struct, e.g. like here https://github.com/gopherjs/vecty/blob/master/examples/todomvc/components/itemview.go#L61

I've extensively looked into the interface-style here, but due to the above issues not having any good solutions, it kills that idea.

**Comment by slimsag** [comment]
> I still have a not so good gut feeling about this change, but let's first iterate on the implementation and then I'll evaluate it again. I'll post some in-line comments.

Just wanted to also say I appreciate your open-minded approach here, Richard 😄 If we can't come up with good logical arguments against this PR I think "I have a gut feeling this change is bad" is totally an okay final verdict which I can accept.

**Comment by influx6** [comment]
@slimsag Sorry for the late reply, just been busy and finally saw the email in my box after so long.

1. How data flows: i.e. how React properties flow from parent components into child components, and how data is accessed externally e.g. like at

I had looked into this problem heavily myself and it really is a pain due to the fact that we do need that flow of control when it comes to initializing pre-built components with data. 
After much deliberation with myself, I scrapped the idea and approach for sticking with a pure Go composition style thinking, which necessary does not compose structs between themselves but leverages the ability for a new component to compose the rendered result of a sub-component. This does alter the flow to a reverse order where its `child to parent` relationship but it resolves the issue and allows the interface approach to trive.

For example in my code piece above, the budget struct has a list of Budget-Items which it then builds off the rendering as it decides.
```
// Budget defines a collection of cost items writting against a given pocket
// budget, it hosts the central items for a budget.
type Budget struct {
	Title            string  `json:"title"`
	Price            float64 `json:"price"`
	items            []BudgetItem
	currency         currency.Currency
	action           int64
	activeBudgetItem int
}
```

Taking that thinking into a more general view, if a master component has a list of subcomponents which are initialized through the parent properties

```
// AddItem adds a new budget item into the lists of Budgets.
func (b *Budget) AddItem(title string, desc string, price float64) {
	bi := BudgetItem{
		Title:  title,
		Desc:   desc,
		Price:  price,
		Time:   time.Now(),
		Budget: b,
	}

	atomic.AddInt64(&b.action, 1)
	{
		b.items = append(b.items, bi)
	}
	atomic.AddInt64(&b.action, -1)
}
```
Then we can equally compose the final form without even touching the base of creating a standard data flow because algorithmically they form such bonds. 
Basically unless you plan to do use alot of reflect magic and interface{} being passed down to subcomponents, let the devs create that relationship between their components and subcomponents.

The beauty of the interface approach is that, it lets them plug into a system that allows reconciliation and rendering well handled for them and reduces your headache because with Go dataflow is a deep rabbit hole with out which a central technique fits.

My principle now simple is: Compose at the struct level, let the program build his components with others at the code level, let them define and design the relationships and data flow in the best idiomatic way possible, by meeting my interface standard I can render and reconcile their views for them.

2. How a component itself requests that its body be reconciled / rerendered upon some state change (e.g. a click event), which prior to this PR is done via an embedded Composite struct and in this PR is done via an embedded Core struct

Simple answer: **Signals**

Infact you have a beautiful signal propagator right now https://github.com/gopherjs/vecty/blob/master/examples/todomvc/components/itemview.go#L67-L70

You are just not using it for that. I adopted this same approach for my view manager:
https://github.com/influx6/gu/blob/master/guviews/views.go#L126-L154
https://github.com/influx6/gu/blob/master/guviews/views.go#L171-L214

Because you have a central backbone system that components can registery to, with a little reflect and tweaking you get something like [Mque](https://github.com/influx6/faux/tree/master/mque), which lets you send any type down the subscriber line and those who match that type get to be called to action.

This allows you to build custom signal structs for different operations, e.g

UpdateView
UpdateStateWithURL
UpdateChildren
UpdateWhenChildChanges

....etc

You get to free your couple and easily have a system in place for notifications.

dom.go
shurcooL: > just because `reflect` seems cheaty / nasty.. but I'd be interested to hear unbiased opinions? @shurcooL can give thoughts too maybe?

I don't know, it doesn't seem that bad to me. It looks like a case where reflect might be a right tool.

Is it guaranteed to have `*Core` or can you only expect that to be the case due to convention?

If there's something to be opposed to, I think it might be the higher level considerations, but IMO it's not the use of reflect for that specific task.

PR #41, opened 2016-04-23
README: add community section

slimsag:


slimsag:
Q: should we use the `#vecty` channel or not split up as much and use `#gophers` channel? I'm not sure which is most appropriate / am okay with whatever (though obviously don't want to derail `#gophers` at all).

shurcooL:
I'd prefer #gopherjs **until** there's so much Vecty-specific discussion that:

- It spams #gopherjs and makes it hard to discuss other things there.
- #vecty won't be too quiet.
- Is too specific to #vecty as to be at all interesting to general #gopherjs.

I'm afraid #vecty would be too quiet. But maybe I'm wrong.

README.md
```diff
@@ -23,3 +23,8 @@ Goals
 	-	Same performance as just using plain JS/HTML/CSS.
 -	Composability
 	-	Nest components to form your entire user interface, seperate them logically as you would any normal Go package.
+
+Community
+=========
+
+- [Join us](https://gophersinvite.herokuapp.com/) in the `#vecty` channel on Gophers Slack!
```
shurcooL: I would change the https://gophersinvite.herokuapp.com link to be over "Gophers Slack" rather than "Join us" text.

You could also linkify the "`#vecty` channel" text with a link to https://gophers.slack.com/messages/vecty/. If you're already logged in to Gophers Slack, you'll end up in the right place. But then make it clear how to get an invite to it if you're not.

shurcooL:
Actually, I didn't realize/forgot that #vecty the channel already exists. In that case, I think linking it it is fine, but maybe also mention #gopherjs channel. Or don't. It's fine. :)

slimsag:
No, I think you're right it would be rather quiet. I do think it would be rather pertinent to gopherjs, but obviously we could also keep gopherjs to be "gopherjs and nothing else" and that'd be OK too. After your explanation, using #gopherjs instead of #vecty makes the most sense to me. I'd rather unite than divide 😄 

PTAL, and I'll obviously wait for @neelance's say before merging this in case he objects.

shurcooL:
I think each bigger channel should be okay to covering a broader range of topics.

E.g., #general is for all Go related discussion, which could even include GopherJS.

#gopherjs is for all GopherJS related discussion, which clearly includes Vecty and anything else.

But if you really want to get into heavy Vecty-specific discussion, then you can be quite sure there won't be non-vecty stuff in #vecty.

neelance:
+1 to all above. LGTM

README.md
```diff
@@ -23,3 +23,8 @@ Goals
 	-	Same performance as just using plain JS/HTML/CSS.
 -	Composability
 	-	Nest components to form your entire user interface, seperate them logically as you would any normal Go package.
+
+Community
+=========
+
+- Join us in the [#gopherjs channel](https://gophers.slack.com/messages/gopherjs/) on the [Gophers Slack!](https://gophersinvite.herokuapp.com/)
```
shurcooL: Minor, I would pull the exclamation point out of the link. Otherwise it feels like it's something called "Slack!" rather than "Slack".

slimsag: Fixed and then squashed.

PR #43, opened 2016-05-06
RenderAsBody: avoid duplicate body in DOM

neelance:
This solution is the only one that I found which does not flicker on refresh and does not result in having two body tags in the DOM.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -19,9 +19,16 @@ func Render(comp Component, container *js.Object) {
 // RenderAsBody renders the given component as the body of the page, replacing
 // whatever existing content in the page body there may be.
 func RenderAsBody(comp Component) {
-	body := js.Global.Get("document").Call("createElement", "body")
+	doc := js.Global.Get("document")
+	body := doc.Call("createElement", "body")
 	Render(comp, body)
-	js.Global.Get("document").Set("body", body)
+	if doc.Get("readyState").String() == "loading" {
+		doc.Call("addEventListener", "DOMContentLoaded", func() { // avoid duplicate body
+			doc.Set("body", body)
+		})
+		return
+	}
+	doc.Set("body", body)
 }
 
 type textComponent struct {


shurcooL:
```
Searching 1228 files for "document.AddEventListener("DOMContentLoaded", false, func("
19 matches across 19 files
```

I have a lot of experience doing this, and I will point out a "gotcha" that you may run into with this.

Depending on packages you import, and what variables you have defined at package level, you may get into a situation where `main` executes so late, that `DOMContentLoading` event gets fired before the listener is added, so the code inside never executes.

It can happen if you (or a package you import) have package-level variables such as:

```Go
var t = template.Must(template.New("").Parse(`...`)
```

Or some `regexp.MustCompile` (especially because of https://github.com/gopherjs/gopherjs/issues/390) package-level variables.

So, I would be careful with this solution.

I have not yet found a bullet-proof way to defer execution after the DOM body exists in situations where package-level variables may cause `main` to run "too late". I welcome suggestions for how to do better.

neelance:
@shurcooL I've changed it a bit, please take another look.

shurcooL:
That looks a lot like a solution that I was considering. I think it will work really well, and it will resolve the issue, so LGTM. I have one minor question I'll post inline.

Because this snippet is something I use so often, I'm still interested in looking for ways to make it better, for example, by making it shorter.

dom.go
```diff
@@ -19,9 +19,16 @@ func Render(comp Component, container *js.Object) {
 // RenderAsBody renders the given component as the body of the page, replacing
 // whatever existing content in the page body there may be.
 func RenderAsBody(comp Component) {
-	body := js.Global.Get("document").Call("createElement", "body")
+	doc := js.Global.Get("document")
+	body := doc.Call("createElement", "body")
 	Render(comp, body)
-	js.Global.Get("document").Set("body", body)
+	if doc.Get("readyState").String() == "loading" {
```
shurcooL: My minor question here is this.

Is there a non-zero possibility of a "race condition here? Imagine the page is loading when the condition on the line is evaluated. Then, it finishes loading. The next 4 lines add a listener which will never execute and return.

shurcooL:
> I can do some local testing in a bit to confirm it works.

I [tested it out locally](https://github.com/shurcooL/resume/commit/a8782b50b9c95d6c4d0615d25096b2da2b22ae11) and it worked as expected.

dom.go
slimsag: I think there is a zero possibility of race condition here, because JS doesn't run async with browser rendering (JS pauses browser rendering). This might not hold 100% true for gopherjs goroutines.. though I'm not sure.

For record, at least jQuery's `$.ready(...)` method is also implemented in this same fashion: https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/core/ready.js#L67-L84

slimsag:
LGTM

dom.go
neelance: Exactly! You can assume that rendering and JS execution do not happen in parallel. Modern browsers do some async tricks, but they take care to not violate this assumption.

shurcooL: Great, thanks for confirming that!

According to the jQuery comment, it seems it may have trouble on IE older than 9, but that's fine with me, I care about supporting only the modern evergreen browsers.

PR #45, opened 2016-06-10
Add StopPropagation for events

davelondon:


shurcooL:
LGTM, but people more familiar with this project should decide if this is in scope.

markup.go
```diff
@@ -113,6 +114,14 @@ func (l *EventListener) PreventDefault() *EventListener {
 	return l
 }
 
+// StopPropagation prevents further propagation of the current event in the capturing and bubbling phases.
```
slimsag: (nit) prefer to wrap this at 80 chars

slimsag:
LGTM except one minor comment

markup.go
shurcooL: Just curious, is that your personal preference, or do you think that it's best practice (listed somewhere else)?

I agree for this comment, because it's top level, that it's nicer to wrap it to 80-100~ish lines (which is optimal for reading).

But there are other comment types that provide very specific detail, like inline notes, that I don't mind keeping very long and single-line. That way they're out of sight except when you really need to read about that particular detail. Here's an [example](https://github.com/shurcooL/markdownfmt/blob/b0700cd85e465d93db670e04763fc6f4070e6e76/markdown/main.go#L379) of what I mean. Disclaimer: I use a computer where I can scroll horizontally as easily as I can scroll vertically.

slimsag: > Just curious, is that your personal preference, or do you think that it's best practice (listed somewhere else)?

> I agree for this comment, because it's top level, that it's nicer to wrap it to 80-100~ish column lines (which is optimal width for reading).

I wouldn't say that "you should always use 80 chars" but I would say "as a general rule of thumb, wrap at 80 chars". In my mind, it doesn't apply to all situations but does to most (e.g. the case here).

> But there are other comment types that provide very specific detail, like inline notes, that I don't mind keeping very long and single-line. That way they're out of sight except when you really need to read about that particular detail. Here's an example of what I mean. Disclaimer: I use a computer where I can scroll horizontally as easily as I can scroll vertically.

For these, I don't have too much preference. I think they are OK off to the side like this, and also OK as comments before that line wrapped at 80 chars. Both are okay in my opinion.

davelondon: I've been using 100 chars as a guide, but I'm perfectly happy with 80. Obviously there's times like [this](https://github.com/kego/ke/blob/2a9338538fc5913b2bb3c986aad826bbe02ea152/system/node/node.go#L24) where breaking a comment would cause gofmt to break the nice indenting.

davelondon:
Done.

neelance:
LGTM

PR #48, opened 2016-06-26
go generate ./elem

slimsag:
Reflects only recent doc changes on https://developer.mozilla.org/en-US/docs/Web/HTML/Element

### DIFF
--- a/elem/elem.gen.go
+++ b/elem/elem.gen.go
@@ -385,7 +385,7 @@ func Italic(markup ...vecty.Markup) *vecty.Element {
 	return e
 }
 
-// The HTML Inline Frame Element (<iframe>) represents a nested browsing context, effectively embedding another HTML page into the current page. In HTML 4.01, a document may contain a head and a body or a head and a frame-set, but not both a body and a frame-set. However, an <iframe> can be used within a normal document body. Each browsing context has its own session history and active document. The browsing context that contains the embedded content is called the parent browsing context. The top-level browsing context (which has no parent) is typically the browser window.
+// The HTML Inline Frame Element (<iframe>) represents a nested browsing context, effectively embedding another HTML page into the current page. In HTML 4.01, a document may contain a head and a body or a head and a frameset, but not both a body and a frameset. However, an <iframe> can be used within a normal document body. Each browsing context has its own session history and active document. The browsing context that contains the embedded content is called the parent browsing context. The top-level browsing context (which has no parent) is typically the browser window.
 //
 // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe
 func InlineFrame(markup ...vecty.Markup) *vecty.Element {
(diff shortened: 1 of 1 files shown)


elem/elem.gen.go
```diff
@@ -754,7 +754,7 @@ func Small(markup ...vecty.Markup) *vecty.Element {
 	return e
 }
 
-// The HTML <source> element specifies multiple media resources for either the <picture>, the <audio> or the <video> element. It is an empty element. It is commonly used to serve the same media content in multiple formats supported by different browsers.
+// Editorial review completed.
```
shurcooL: Doesn't this look like a regression? :/

shurcooL:
Some of the changes are weird, but, that's what the upstream is. ¯\_(ツ)_/¯

LGTM.

elem/elem.gen.go
slimsag: It does.. good catch. It appears this is what is one the page at https://developer.mozilla.org/en-US/docs/Web/HTML/Element#Embedded_content though.

The top of the page says this after shift+refresh:

image

Perhaps it will update sometime soon, but history has told me it might be a while (on the order of days).

elem/elem.gen.go
```diff
@@ -889,7 +889,7 @@ func TableHead(markup ...vecty.Markup) *vecty.Element {
 	return e
 }
 
-// The HTML <time> element represents either a time on a 24-hour clock or a precise date in the Gregorian calendar (with optional time and timezone information).
+// Technical review completed.
```
slimsag: same thing here I guess

elem/elem.gen.go
```diff
@@ -970,7 +970,7 @@ func WordBreakOpportunity(markup ...vecty.Markup) *vecty.Element {
 	return e
 }
 
-// Heading elements implement six levels of document headings, <h1> is the most important and <h6> is the least. A heading element briefly describes the topic of the section it introduces. Heading information may be used by user agents, for example, to construct a table of contents for a document automatically.
+// Heading elements implement six levels of document headings, <h1> is the most important and <h6> is the least. A heading element briefly describes the topic of the section it introduces. Heading information may be used by user agents, for example, to construct a table of contents for a document automatically( just like the fixed sider bar of this page on the right).
```
slimsag: I've fixed the spacing issue around the `()` brackets in the MDN docs, no clue when they will update.

neelance:
Can we fix the issues upstream?

slimsag:
@neelance that's exactly what I am trying to do :)

elem/elem.gen.go
slimsag: The upstream page isn't updating.

- Fixed on the source page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements 
- Won't update on the primary page, ctrl+f for `( just like`, https://developer.mozilla.org/en-US/docs/Web/HTML/Element
- Doing shift+refresh will queue the page for an update.. but even after finished it just doesn't update.

elem/elem.gen.go
slimsag: Shame, the `<source>` tag still has a description of `Editorial review completed.`.

It seems like the primary page won't rebuild for some reason.

elem/elem.gen.go
slimsag: Nevermind; I've fixed the `( just like` spacing issue and also reverted the `( just like the fixed sider bar of this page on the right).` change entirely in the MDN since it makes the source page content brittle to layout changes on MDN (and had a spacing issue, and a typo).

elem/elem.gen.go
slimsag: Fixed in the MDN source too.

elem/elem.gen.go
slimsag: Fixed this in the MDN source, turned out to be due to incorrect ordering of the page elements do the description was borked.

slimsag:
Okay, fixed all of the issues in the source after an hour or so of experimentation on their site. PTAL @shurcooL @neelance

shurcooL:
LGTM.

slimsag:
I expect that this change won't be controversial to @neelance now that it is just improvements; so merging.

PR #51, opened 2016-06-27
go generate ./event

slimsag:


slimsag:
PTAL @neelance @shurcooL

event/generate.go
```diff
@@ -116,6 +116,9 @@ func main() {
 		"visibilitychange":         "VisibilityChange",
```
shurcooL: This is out of scope, but it'd be nice to document `nameMap` to explain what its purpose is. It's easy to forget...

IIRC, it's to generate Go names that are compatible with Go style on [MixedCaps](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps) and [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). But I can't remember if that's exactly it, or there's another reason for this.

event/event.gen.go
```diff
@@ -1092,6 +1092,27 @@ func UserProximity(listener func(*vecty.Event)) *vecty.EventListener {
...
+func VRDisplayDisconnected(listener func(*vecty.Event)) *vecty.EventListener {
+	return &vecty.EventListener{Name: "vrdisplaydisconnected", Listener: listener}
+}
+
+// The presenting state of a VR device has changed — i.e. from presenting to not presenting, or vice versa.
+//
+// https://developer.mozilla.org/docs/Web/Events/vrdisplaypresentchange
+func VRDisplayPresentChange(listener func(*vecty.Event)) *vecty.EventListener {
+	return &vecty.EventListener{Name: "vrdisplaypresentchange", Listener: listener}
+}
```
shurcooL: FWIW, all these comments break the style rule that documentation should begin with the name of the identifier, e.g.:

```Go
// VRDisplayPresentChange ...
func VRDisplayPresentChange() {
```

And it's something `go lint` detects.

Since this documentation is generated from an upstream source that doesn't follow that rule, I'm not sure if anything can be done about it.

shurcooL:
This is mostly a generated change, do these really need to be reviewed?

LGTM.

event/event.gen.go
```diff
@@ -1029,20 +1029,6 @@ func TouchEnd(listener func(*vecty.Event)) *vecty.EventListener {
...
-	return &vecty.EventListener{Name: "touchenter", Listener: listener}
-}
-
-// A touch point is moved off the interactive area of an element.
-//
-// https://developer.mozilla.org/docs/Web/Events/touchleave
-func TouchLeave(listener func(*vecty.Event)) *vecty.EventListener {
-	return &vecty.EventListener{Name: "touchleave", Listener: listener}
-}
-
```
slimsag: FYI these `TouchEnter` and `TouchLeave` events are removed because they were removed from the actual spec and were never really implemented.

event/event.gen.go
slimsag: Yes, this is a good point. I think we can solve this with some slight management / text mingling within the generator because they all follow the same paragraph style. I'll see if I can fix this in a future PR.

event/generate.go
slimsag: Agree, I'll send this as part of a future PR.

slimsag:
> This is mostly a generated change, do these really need to be reviewed?

I think you make a good point, I might've been too OCD here :) I'm going to merge now and in the future I will self-review these and commit directly when they appear to be plain/simple.

PR #52, opened 2016-07-03
General improvements to generated documentation

slimsag:
- Keep comments in both `event` and `elem` packages at readable lengths (wrap at 80 chars).
- Start func docstrings with the func name itself (following Go style conventions).
- Document name mapping properly (/cc @shurcooL).

### DIFF
--- a/event/generate.go
+++ b/event/generate.go
@@ -19,6 +19,12 @@ type Event struct {
 }
 
 func main() {
+	// nameMap translates lowercase HTML attribute names from the MDN source
+	// into a proper Go style name with MixedCaps and initialisms:
+	//
+	//  https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps
+	//  https://github.com/golang/go/wiki/CodeReviewComments#initialisms
+	//
 	nameMap := map[string]string{
 		"afterprint":               "AfterPrint",
 		"animationend":             "AnimationEnd",
--- a/event/generate.go
+++ b/event/generate.go
@@ -153,9 +153,7 @@ func main() {
 		}
 
 		if e.Desc != "" {
-			// Lowercase the first letter of the description.
-			lowercaseDesc := strings.ToLower(string(e.Desc[0])) + e.Desc[1:]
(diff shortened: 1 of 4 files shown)


shurcooL:
This looks incredible! Thanks for doing it.

> Start func docstrings with the func name itself (following Go style conventions).

Are you looking to cover all cases, or just the vast majority? I see some entries that would still trigger golint warnings, would you want me to point them out?

slimsag:
> Are you looking to cover all cases, or just the vast majority? I see some entries that would still trigger golint warnings, would you want me to point them out?

Correct, there are still a large number that do cause golint warnings at this point (I would estimate 20-30% but don't have a real number on hand); These will mostly be solved by updating the MDN and regenerating -- which can be done independently of this PR. I will follow-up on these :)

event/generate.go
```diff
@@ -138,15 +145,20 @@ func main() {
...
+		e.Spec = strings.TrimSpace(cols.Eq(2).Text())
 
 		funName := nameMap[e.Name]
 		if funName == "" {
 			funName = capitalize(e.Name)
 		}
 
+		if e.Desc != "" {
+			// Lowercase the first letter of the description.
+			lowercaseDesc := strings.ToLower(string(e.Desc[0])) + e.Desc[1:]
```
shurcooL: It seems inconsistent that `capitalize` is also used in just one place, yet it's made into a helper, but the same isn't done for its lower equivalent.

shurcooL: As a bonus, if you make it an unexported helper, then the comment above can be removed since it'd be unneccessary.

shurcooL:
One minor comment. LGTM.

event/generate.go
slimsag: Very good suggestion, thanks for this! Done. :)

slimsag:
Since there have been no further comments and I've addressed all by shurcooL, I'm going to merge. I think this change won't be very controversial, but @neelance if you see anything here you dislike or want to discuss we can revert that portion and I'll resubmit it as a PR for further discussion.

PR #62, opened 2016-10-31
(final version) new component architecture

### LINKED ISSUE #27: document how to skip/short-circuit reconciliation
If e.g. you are certain that the body / what `Render` returns will not have changed.

slimsag:
Some months ago I wrote several component architecture versions, this ended up being the best out of all of them (including the current one we have). I intend to move forward with this implementation, but welcome direct feedback on how to improve it. Let's make great improvements to Vecty and get this project usable (but let's not hinder ourselves on communication or theoretical ideas).

Inline comments show the things you should look at. Ignore generated files `elem/` and `style/` folders.

Fixes #27

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -74,13 +74,17 @@ type HTML struct {
 	children        []ComponentOrHTML
 }
 
-func (h *HTML) restoreHTML(prev *HTML) {
+func (h *HTML) restoreText(prev *HTML) {
 	h.Node = prev.Node
 
 	// Text modifications.
 	if h.text != prev.text {
 		h.Node.Set("nodeValue", h.text)
 	}
+}
+
+func (h *HTML) restoreHTML(prev *HTML) {
+	h.Node = prev.Node
 
 	// Properties
 	for name, value := range h.properties {
@@ -169,8 +173,14 @@ func (h *HTML) Restore(old ComponentOrHTML) {
 	}
 
 	if prev, ok := old.(*HTML); ok && prev != nil {
-		h.restoreHTML(prev)
-		return
+		if h.text != "" && prev.text != "" {
+			h.restoreText(prev)
+			return
+		}
+		if h.tag != "" && prev.tag != "" && h.tag == prev.tag {
+			h.restoreHTML(prev)
+			return
+		}
 	}
 
 	if h.tag != "" && h.text != "" {
--- a/dom.go
+++ b/dom.go
@@ -1,92 +1,162 @@
 package vecty
 
-import "github.com/gopherjs/gopherjs/js"
+import (
+	"fmt"
 
-// Component represents a Vecty component.
+	"github.com/gopherjs/gopherjs/js"
+)
+
+// Core implements the Context method of the Component interface, and is the
+// core/central struct which all Component implementations should embed.
+type Core struct {
+	prevRender *HTML
+}
+
+// Context implements the Component interface.
+func (c *Core) Context() *Core { return c }
+
+// Component represents a single visual component within an application. To
+// define a new component simply implement the Render method and embed the Core
+// struct:
+//
+// 	type MyComponent struct {
+// 		vecty.Core
+// 		... additional component fields (state or properties) ...
+// 	}
+//
+// 	func (c *MyComponent) Render() *vecty.HTML {
+// 		... rendering ...
+// 	}
+//
 type Component interface {
-	Markup
-	Reconcile(oldComp Component)
-	Node() *js.Object
+	// Render is responsible for building HTML which represents the component.
+	Render() *HTML
+
+	// Context returns the components context, which is used internally by
+	// Vecty in order to store the previous component render for diffing.
+	Context() *Core
 }
 
-// Render renders a component into the given container element. It is appended
-// as a child element.
-func Render(comp Component, container *js.Object) {
-	comp.Reconcile(nil)
-	container.Call("appendChild", comp.Node())
+// ComponentOrHTML represents one of:
+//
+//  Component
+//  *HTML
+//
+// If the underlying value is not one of these types, the code handling the
+// value is expected to panic.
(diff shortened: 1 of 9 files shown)


dom.go
```diff
@@ -1,92 +1,164 @@
...
+	// previous instance of a component. The previous component may be nil or
+	// of a different type than this Restorer itself, thus a type assertion
+	// should be used.
+	//
+	// If skip = true is returned, restoration of this component's body is
+	// skipped. That is, the component is not rerendered. If the component can
+	// prove when Restore is called that the HTML rendered by Component.Render
+	// would not change, true should be returned.
+	Restore(prev Component) (skip bool)
 }
```
slimsag: - `Reconcile` is now named `Restore`. In terms of English, `Reconcile` is quite the loaded word to spell and `Restore` has the same meaning. I think this will be a win for people not used to React wording, but a small loss for those who are used to React wording. Simple naming == better naming, so we'll go with this.
- How to short-circuit component restoration is now documented :) Fixes #27

dom.go
```diff
@@ -1,92 +1,164 @@
...
+//
+// 	type MyComponent struct {
+// 		vecty.Core
+// 		... additional component fields (state or properties) ...
+// 	}
+//
+// 	func (c *MyComponent) Render() *vecty.HTML {
+// 		... rendering ...
+// 	}
+//
```
slimsag: Component creation is greatly simplified (and documented) with this new pattern. `Apply` is no longer needed for components (which was confusing from a user perspective), and `Restore` is optional / a default is provided. It means you only need to use what you want.

dom.go
```diff
@@ -1,92 +1,164 @@
...
-		return
-	}
+// ComponentOrHTML represents one of:
+//
+//  Component
+//  *HTML
+//
+// If the underlying value is not one of these types, the code handling the
+// value is expected to panic.
+type ComponentOrHTML interface{}
```
slimsag: Union-like types are used to differentiate between HTML (`DOMNode` or `TextNode`) and components (user built components). This way `*HTML` (previously `Element`) can distinguish the two and straightforward decisions about where the previous render is etc.

examples/todomvc/components/itemview.go
```diff
@@ -52,25 +45,25 @@ func (p *ItemView) onToggleCompleted(event *vecty.Event) {
 func (p *ItemView) onStartEdit(event *vecty.Event) {
 	p.editing = true
 	p.editTitle = p.Item.Title
-	p.ReconcileBody()
-	p.input.Node().Call("focus")
+	vecty.Rerender(p)
```
slimsag: Instead of components calling `(*MyComponent).Composite.ReconcileBody()`, they just call `vecty.Rerender(myComponent)`.

While `Rerender` uses the component's `vecty.Core` embedded struct internally, the overall end-user perception is much more straightforward. Also `Rerender` sounds simpler / more understandable than `ReconcileBody` (easy to confused with the DOM `<body>` tag).

examples/todomvc/components/itemview.go
```diff
@@ -52,25 +45,25 @@ func (p *ItemView) onToggleCompleted(event *vecty.Event) {
...
-	p.ReconcileBody()
+	vecty.Rerender(p)
 	dispatcher.Dispatch(&actions.SetTitle{
 		Index: p.Index,
 		Title: p.editTitle,
 	})
 }
 
-func (p *ItemView) render() vecty.Component {
+func (p *ItemView) Render() *vecty.HTML {
```
slimsag: Component rendering is now done via implementing a public `Render` method which returns `*vecty.HTML`.

Simple, and it makes sense. Beginners do not ask the question "Why does my component render another component..?". Components can now be thought of as a way of grouping HTML elements, we no longer mix the concepts of elements and components together.

dom.go
shurcooL: I think the two words are equally easy to spell (to me), but restore is indeed a simpler (less fancy) word. If what this does is "restore", then it's definitely a better name (for those not already familiar with React).

One comment, I didn't realize they had the same meaning. To me, "reconcile" means something like "to cause to coexist in harmony" or "to resolve some sort of differences". But restore means to take something and make it like it used to be (restore to a previous state).

So I want to ask, is that really true? Does/did reconcile do nothing more than restore to a previous state?

davelondon:
Hey @slimsag I found a couple of bugs and sent PRs to this branch: https://github.com/gopherjs/vecty/pull/63 and https://github.com/gopherjs/vecty/pull/64

I'm still getting my head around the changes, but it seems really good so far.

davelondon:
One thing that's been causing more problems today... It would be great if we could handle `nil` in a more natural way to indicate nothing should be rendered.

1) As the result of a render function:

```
func (v *MyView) Render() *vecty.HTML {
	if v.Foo == nil {
		return nil
	}
	return elem.Div(
		...
	)
}
```

2) As markup:

```
func (v *MyView) Render() *vecty.HTML {
	var foo *vecty.HTML
	if ... {
		foo = elem.Div(vecty.Text("foo"))
	}
	return elem.Div(
		foo
	)
}
```

I'm still trying to work out the ramifications of this... It's not simple. I believe prior to this PR, returning `nil` from a component render function was not supported, but using nil as markup was OK... Now it seems both scenarios cause problems.

It's something that can be worked around, but it would be much nicer if both were supported... What do you think?

slimsag:
@davelondon I completely agree, we need some indicator like `nil` to suggest "do not render anything". I haven't looked 100% extensively into this, but I think we should support it nicely somehow.

Let's do a prompt follow-up after this PR :) If you've already thought out most edge cases I'd be happy to review, otherwise I'll start digging into it soon

slimsag:
Oh I see I misread your PR comment, this is a regression from previous behavior because now you have no way to do that. Got it.

I'll work on implementing option #1 you presented above. I think that makes sense.

PR #63, opened 2016-11-12
Initialise dataset if it's nil

davelondon:
I think we need to initialise h.dataset here...


slimsag:
Yep, you're right. Thanks for this!

PR #64, opened 2016-11-12
Correctly restore mixed text/tag elements

davelondon:
I believe restoreHTML should only be called when both elements are either:

* both text
...or...
* both tags and have the same tag

Is that right? I found that splitting restoreHTML into restoreHTML and restoreText made it clearer, but this isn't essential...

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -74,13 +74,17 @@ type HTML struct {
 	children        []ComponentOrHTML
 }
 
-func (h *HTML) restoreHTML(prev *HTML) {
+func (h *HTML) restoreText(prev *HTML) {
 	h.Node = prev.Node
 
 	// Text modifications.
 	if h.text != prev.text {
 		h.Node.Set("nodeValue", h.text)
 	}
+}
+
+func (h *HTML) restoreHTML(prev *HTML) {
+	h.Node = prev.Node
 
 	// Properties
 	for name, value := range h.properties {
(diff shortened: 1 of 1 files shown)


slimsag:
You are correct. Bear in mind I wrote this code some ~6 months ago 😃 so I had to triple check to confirm.

- Everything in `restoreHTML` after this change is indeed non-applicable to text nodes, including event listeners.
- Your guess about only wanting to restore two `*HTML` when they are both either a `TextNode` or same tag is correct.

dom.go
```diff
@@ -169,8 +173,14 @@ func (h *HTML) Restore(old ComponentOrHTML) {
...
-		h.restoreHTML(prev)
-		return
+		if h.text != "" && prev.text != "" {
+			h.restoreText(prev)
+			return
+		}
+		if h.tag != "" && prev.tag != "" && h.tag == prev.tag {
+			h.restoreHTML(prev)
+			return
+		}
```
slimsag: Also this would be written more straightforward like this I think:

```
if h.text != "" && prev.text != "" {
	// Two text nodes.
	h.restoreText(prev)
	return
} else if h.tag == prev.tag {
	// Two HTML elements of the same tag.
	h.restoreHTML(prev)
	return
}
```

I'll make this change after merging.

davelondon: If both `h` and `prev` are text nodes containing an empty string, we probably shouldn't call `restoreHTML`... Not sure what you think of empty string text nodes, but they're possible to create with `vecty.Text("")`, so we should watch out for them (see https://github.com/gopherjs/vecty/pull/67)

slimsag: Oh I missed this comment (GitHub didn't auto refresh..) you're right. I will revert my suggested change.

PR #67, opened 2016-11-12
Support text nodes with an empty string

davelondon:
Should we support text nodes with an empty string? I had an empty text node that was causing problems...


slimsag:
Thanks! LGTM

PR #68, opened 2016-11-12
Better support for nil in apply

davelondon:
If `m` is a `(*HTML)(nil)` then the first nil check will fail. This adds another after the cast to *HTML.


slimsag:
Perfect, thanks for this!

PR #69, opened 2016-11-13
Add optional Unmounter interface that Components may implement.

### LINKED ISSUE #58: Element-level Mount and Unmount events
We should support giving elements (not components) a callback when the element is mounted into the DOM / unmounted from the DOM.

slimsag:
Fixes #58

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -37,6 +37,14 @@ type Component interface {
 	Context() *Core
 }
 
+// Unmounter is an optional interface that a Component can implement in order
+// to receive component unmount events.
+type Unmounter interface {
+	// Unmount is called after the component has been unmounted, in a separate
+	// goroutine, after the DOM element has been removed.
+	Unmount()
+}
+
 // ComponentOrHTML represents one of:
 //
 //  Component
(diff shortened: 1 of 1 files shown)


davelondon:
I'm not sure... would be great if it does because it's super simple. Is forcing it into another goroutine essential? Seems like if the end used wanted it in a goroutine they could spawn one themselves, but not the other way around... I'm just working through this restoreHTML bug then I'll have a good look at this.

slimsag:
re: goroutine: good point.. my thinking was that forcing it to be in a separate goroutine meant that the UI 'thread' would not be blocked by users doing significant work as part of `Unmount` -- but I'm actually now thinking this is a bad thing esp. with your above comment. I'll update this PR to not use goroutines like this.

slimsag:
I've removed the goroutine portion of this, as said earlier. I think this implementation will work OK for most situations; we can figure out any issues that arise later and move forward with this for now :)

dom.go
```diff
@@ -37,6 +37,14 @@ type Component interface {
 	Context() *Core
 }
 
+// Unmounter is an optional interface that a Component can implement in order
+// to receive component unmount events.
+type Unmounter interface {
+	// Unmount is called after the component has been unmounted, in a separate
+	// goroutine, after the DOM element has been removed.
```
shurcooL: > I've removed the goroutine portion of this

Did you forget the documentation bit here?

slimsag: Indeed I did! 😨 Thanks for pointing this out! 😃 

Fixed in becba2706b0445067f6cc51d9975846d946c8f6c

shurcooL: You're welcome! :)

PR #71, opened 2016-11-15
Prevent reuse of range variable in closure

davelondon:
Prevent reuse of range variable in closure

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -161,7 +161,8 @@ func (h *HTML) restoreHTML(prev *HTML) {
 
 // Restore implements the Restorer interface.
 func (h *HTML) Restore(old ComponentOrHTML) {
-	for _, l := range h.eventListeners {
+	for _, lrange := range h.eventListeners {
+		l := lrange
 		l.wrapper = func(jsEvent *js.Object) {
 			if l.callPreventDefault {
 				jsEvent.Call("preventDefault")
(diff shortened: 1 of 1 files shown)


dom.go
```diff
@@ -161,7 +161,8 @@ func (h *HTML) restoreHTML(prev *HTML) {
 
 // Restore implements the Restorer interface.
 func (h *HTML) Restore(old ComponentOrHTML) {
-	for _, l := range h.eventListeners {
+	for _, lrange := range h.eventListeners {
+		l := lrange
```
shurcooL: Why not just `l := l`? It's a well known pattern, and the idiomatic way of doing this.

> It may seem odd to write
>
> ```Go
> req := req
> ```
>
> but it's legal and idiomatic in Go to do this. You get a fresh version of the variable with the same name, deliberately shadowing the loop variable locally but unique to each goroutine.

Source: https://golang.org/doc/effective_go.html#channels

Also, `lrange` violates the [Mixed Caps](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps) naming convention, it would be `lRange`.

shurcooL:
One style suggestion, but otherwise this is a nice and valid fix. LGTM.

dom.go
davelondon: Yup, done.

slimsag:
Sorry I missed this! LGTM

PR #82, opened 2017-01-30
Now it can deal with multiple style property values to allow vendor-prefixed values

thearchitect:
How styles handling implemented now makes impossible to use vendor-prefixed property values.

For example
```go
elem.Div(
	vecty.Style("display", "-ms-flexbox"),
	vecty.Style("display", "-webkit-box"),
	vecty.Style("display", "-moz-box"),
	vecty.Style("display", "-ms-box"),
	vecty.Style("display", "box"),
)
```
will try to set only `display: box`.

If we want to completely eliminate necessity of writing CSS files we need to handle such cases, 'cause CSS prefixes is very common.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -2,6 +2,7 @@ package vecty
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/gopherjs/gopherjs/js"
 )
@@ -80,6 +81,7 @@ type HTML struct {
 	Node *js.Object
 
 	tag, text, innerHTML   string
+	classes                map[string]bool
 	styles                 map[string]map[string]bool
 	dataset                map[string]string
 	properties, attributes map[string]interface{}
@@ -131,6 +133,8 @@ func (h *HTML) restoreHTML(prev *HTML) {
 			h.Node.Call("removeAttribute", name)
 		}
 	}
+
+	h.populateClassNameProperty()
 
 	// Styles
 	style := h.Node.Get("style")
@@ -258,6 +262,7 @@ func (h *HTML) Restore(old ComponentOrHTML) {
 	for name, value := range h.dataset {
 		dataset.Set(name, value)
 	}
+	h.populateClassNameProperty()
 	style := h.Node.Get("style")
 	for name, value := range h.styles {
 		for value, _ := range value {
@@ -279,6 +284,18 @@ func (h *HTML) Restore(old ComponentOrHTML) {
 			continue
 		}
 		h.Node.Call("appendChild", nextChildRender.Node)
+	}
+}
+
+func (h *HTML) populateClassNameProperty() {
+	var classes []string
+	for class, set := range h.classes {
+		if set {
+			classes = append(classes, class)
+		}
+	}
+	if len(classes) > 0 {
+		h.Node.Set("className", strings.Join(classes, " "))
 	}
 }
 
--- a/dom.go
+++ b/dom.go
@@ -299,6 +299,14 @@ func (h *HTML) populateClassNameProperty() {
 	}
 }
 
+func (h *HTML) Add(m ...MarkupOrComponentOrHTML) *HTML {
+	for _, m := range m {
+		apply(m, h)
+	}
+
+	return h
+}
+
 // Tag returns an HTML element with the given tag name. Generally, this
 // function is not used directly but rather the elem subpackage (which is type
 // safe) is used instead.
--- a/dom.go
+++ b/dom.go
@@ -80,7 +80,8 @@ type HTML struct {
 	Node *js.Object
 
 	tag, text, innerHTML   string
-	styles, dataset        map[string]string
+	styles                 map[string]map[string]bool
+	dataset                map[string]string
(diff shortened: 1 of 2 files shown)


dom.go
```diff
@@ -80,7 +80,8 @@ type HTML struct {
 	Node *js.Object
 
 	tag, text, innerHTML   string
-	styles, dataset        map[string]string
+	styles                 map[string]map[string]bool
```
gernest: Why mo just use `map[string]ClassMap`? it looks like ClassMap can handle multiple classes already! 

I was also thinking about something like this!

slimsag:
Sorry for the late review @thearchitect -- thanks for sending a PR! A couple of thoughts:

1. All three of your commits seem entirely unrelated to one another, if I am correct could you please send those as individual PRs instead? That would make it much easier to review and ensure we have a high code quality 😄 
2. It seems like all of your commits don't clearly explain why the change is made, it's generally good practice to explain that in the commit messages or PR.

I'll leave some more concrete comments inline.

dom.go
```diff
@@ -299,6 +299,14 @@ func (h *HTML) populateClassNameProperty() {
 	}
 }
 
+func (h *HTML) Add(m ...MarkupOrComponentOrHTML) *HTML {
+	for _, m := range m {
+		apply(m, h)
+	}
+
+	return h
+}
```
slimsag: It's not clear to me why this is needed or desired? What problem does this solve? Please create an issue and I can fix it or send this as a separate PR and we can discuss there.

dom.go
```diff
@@ -258,6 +262,7 @@ func (h *HTML) Restore(old ComponentOrHTML) {
 	for name, value := range h.dataset {
 		dataset.Set(name, value)
 	}
+	h.populateClassNameProperty()
```
slimsag: In this situation, the behavior should just be inlined (this is element creation, after all)

dom.go
```diff
@@ -132,6 +134,8 @@ func (h *HTML) restoreHTML(prev *HTML) {
 		}
 	}
 
+	h.populateClassNameProperty()
```
slimsag: In this situation, I believe the behavior isn't correct. This would cause us to set `className` on the node every single time anything on the element is changed -- regardless of whether or not `className` itself has actually changed since the last render. That violates the standard reconciliation pattern being used here.

What is the motivation for moving `className` property setting into `*HTML` as if it's a first-class thing? After all, `className` is just a standard property like everything else -- there is no need for us to do this here AFAIK.

markup.go
```diff
@@ -144,13 +143,24 @@ type ClassMap map[string]bool
...
-		if active {
-			classes = append(classes, name)
-		}
+		h.classes[name] = active
 	}
-	Property("className", strings.Join(classes, " ")).Apply(h)
+}
+
+// Class returns markup that applies all given classes to an element
+func Class(classes ...string) Markup {
```
slimsag: Overall I like this change, but it should be implemented differently. We don't need `className` to be a part of `*HTML` as I said in my other comment above; I will send a separate PR to implement this now (this change also should not be a part of this PR)

dom.go
```diff
@@ -133,10 +134,25 @@ func (h *HTML) restoreHTML(prev *HTML) {
 
 	// Styles
 	style := h.Node.Get("style")
-	for name, value := range h.styles {
-		oldValue := prev.styles[name]
-		if value != oldValue {
-			style.Call("setProperty", name, value)
+	for propName, prop := range h.styles {
+		setProperty := func() {
```
slimsag: Generally speaking avoid helpers like these in this function at all costs. They make things arbitrarily hard to reason about.

dom.go
```diff
@@ -244,7 +260,9 @@ func (h *HTML) Restore(old ComponentOrHTML) {
 	}
 	style := h.Node.Get("style")
 	for name, value := range h.styles {
-		style.Call("setProperty", name, value)
+		for value, _ := range value {
+			style.Call("setProperty", name, value)
+		}
```
slimsag: I don't understand this change. This causes us to call `setProperty` with the same `name` but different `value` multiple times?

slimsag: OK; I see now I was wrong about this due to the example in your PR description.

slimsag:
From what I can tell, your idea of invoking `node.style.setProperty` with multiple vendor prefix values will not work. For example try it in a Chrome console:

image

As you would expect from looking at the `setProperty` API, it replaces the last value for the name (`display`).

slimsag:
Here is an exactly identical issue on stackoverflow speaking about how to set this given the name `display` is the same, with a solution: http://stackoverflow.com/questions/17896846/set-display-property-flex-on-element-cross-browser

thearchitect:
Hi slimsag! Thank you for your attention. I've added second an third commits to PR by mistake. I'm going to fix it and address your comments within 2-3 days.

slimsag:
Looks like this PR is a bit inactive, feel free to resubmit whenever (or I'll try to get to it as soon as I have time). Closing for now :)

PR #87, opened 2017-02-26
Fix wrong interface name in panic message.

shurcooL:
Seeing the reference to MarkupOrComponent interface made me look for it,
but it doesn't exist. It's just a typo, originally introduced in 6d402b7baf023b2977854d6345400713ff82fcde. /cc @slimsag


slimsag:
Nice find! Thanks :)

PR #89, opened 2017-02-26
elem: Rename h1-h6 elements from Header to Heading.

### LINKED ISSUE #88: Heading elements (h1, h2, h3, h4, h5, h6) are named Header.
I believe the correct name for https://godoc.org/github.com/gopherjs/vecty/elem#Header1 should be `Heading1`, not `Header1`. Same for 2, 3, 4, 5, 6.

The documentation correctly refers to them as heading elements.

> Head**ing** elements implement six levels of document headings, \<h1\> is the most important and \<h6\> is the least. A heading element briefly describes the topic of the section it introduces. Heading information may be used by user agents, for example, to construct a table of contents for a document automatically.
>
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

I've seen a similar issue happen in https://github.com/russross/blackfriday/issues/330.

An alternative naming that would also be valid is `H1`, etc.

shurcooL:
This PR includes a breaking API change.

- Fix elem package generate script to work with latest changes to https://developer.mozilla.org/en-US/docs/Web/HTML/Element page.

- Regenerate entire elem package.

- Rename h1-h6 elements from Header to Heading (breaking API change).

- Update todomvc example to use new name.

Fixes #88.

The PR is split into 2 logical commits for easier review.


elem/elem.gen.go
```diff
@@ -745,48 +725,41 @@ func Sample(markup ...vecty.MarkupOrComponentOrHTML) *vecty.HTML {
...
-// ordinary HTML. It is used with Web Components.
+// Slot is a placeholder inside a web component that you can fill with your own
+// markup, with the effect of composing different DOM trees together.
 //
-// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Shadow
-func Shadow(markup ...vecty.MarkupOrComponentOrHTML) *vecty.HTML {
-	return vecty.Tag("shadow", markup...)
+// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot
+func Slot(markup ...vecty.MarkupOrComponentOrHTML) *vecty.HTML {
+	return vecty.Tag("slot", markup...)
```
slimsag: Hmm, I wonder if these got removed intentionally.

slimsag: OK, I see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Shadow is obsolete and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot is experimental so this is correct.

slimsag:
LGTM -- thanks for this @shurcooL !

PR #91, opened 2017-02-28
Use canonical format for build constraint.

shurcooL:
This improves consistency.

See https://godoc.org/go/build#hdr-Build_Constraints for definition of a build constraint:

> A build constraint, also known as a build tag, is a line comment that begins
>
>     // +build

### DIFF
--- a/gopherjs_only.go
+++ b/gopherjs_only.go
@@ -1,4 +1,4 @@
-//+build !js
+// +build !js
 
 package vecty
 


shurcooL:
Friendly ping @slimsag, this is a 1 character PR that should be easy to review.

slimsag:
Sorry! I missed this! Thanks for the ping :)

Issue #85, opened 2017-02-12
Incorrect Unmounter behavior in certain situations


slimsag:
> I believe the current behaviour for `Unmounter` is incorrect.  In the case where two components return the same HTML tag, the previous component will be replaced without `Unmount` being called.
>
> Only way I can see to produce the correct behaviour here is via reflection.  I've implemented this in 3df8654e9fac557d5707befd718432533ad7e8c3, and I'm now seeing consistent Mount/Unmount behaviour when I'd expect it.

Issue #75, opened 2017-01-21
Support for fine-grained internal event handling (element creation, re-creation, removal)


pdf:
Finding vecty pretty interesting, just working through a PoC to get my head around it, and one thing I'm butting up against is trying to integrate with 3rd-party JavaScript that expects to have access to elements in the DOM.  There doesn't currently appear to be any way to know when elements are actually rendered.

We have the `Unmount` handler, which is great for one direction, but what about support for a `Mounted` event?  Or am I missing something and thinking about this all wrong?  I'm sure there are some caveats to doing DOM manipulation outside the framework, as elements may get replaced, etc, but there are times when you really need to get at the rendered element.

slimsag:
HI @pdf ! First, cool username :)

Second, yes, I've been aware of this issue for a while and agree a Mounter interface would resolve it. If you don't mind, I'd like to get into the nitty-gritty details of why you need this. For example, could you point me to the exact JS library / relevant code that you would need to interoperate with?

There are absolutely caveats to working with Vecty on an element, e.g. the case of elements being replaced as you already stated. In your situation, would you have to deal with that element-replacement situation?

I'm OK with adding a Mounter interface to get access to the underlying element. In fact I will send a PR for this now, but I don't want to merge it unless we have a really solid real-world use case that can't be solved some other way more cleanly. (I want to avoid the React situation of having a million different *Mount *Unmount handlers that have subtle differences)

pdf:
The first example I encountered this with was looking at bringing in some UI enhancements, like [Materialize](http://materializecss.com/) or [Material Components](https://github.com/material-components/material-components-web/), both of which expect to enhance the DOM with JS.

I have it working in a rudimentary fashion via a very naive first pass at `Mounter` in 462e33816ed2ee5be34b8b7586ba9d0b8f88aeca , though I need to look at it in more detail, especially into how to handle replacement behaviour - I basically just slapped the `Mount` call everywhere the DOM was updated to get up and running, though this is almost certainly not correct.

Material Components, for example, requires calling a `destroy` function when the element is removed from the DOM, which makes me think maybe we'll be leaking something if replace is performed without `Unmount`, though I haven't looked into exactly what they're doing in `destroy`.

Hitting replacements seems pretty certain when dealing with dynamic lists and the like from my brief look at the vecty code, but I just started to have a poke at GopherJS/vecty yesterday, so I'm still in exploration mode.

pdf:
[...]

PR #93, opened 2017-03-14
Implement Mounter, trigger mount/unmount based on component equality

pdf:
I realize #75 and #85 were split out, but the implementation is really intrinsically linked.  Since the issues appear to be stagnating somewhat, I thought I'd submit a PR to see if we can get some forward-progress here.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -186,9 +186,7 @@ func (h *HTML) restoreHTML(prev *HTML) {
 			prevChildRender = prevChild.(Component).Context().prevRender
 		}
 		removeNode(prevChildRender.Node)
-		if u, ok := prevChild.(Unmounter); ok {
-			u.Unmount()
-		}
+		doUnmount(prevChild)
 	}
 }
 
@@ -367,4 +365,11 @@ func AddStylesheet(url string) {
 	link.Set("rel", "stylesheet")
 	link.Set("href", url)
 	js.Global.Get("document").Get("head").Call("appendChild", link)
+}
+
+// doUnmount calls the Unmount function on Unmounter components
+func doUnmount(h ComponentOrHTML) {
+	if u, ok := h.(Unmounter); ok {
+		u.Unmount()
+	}
 }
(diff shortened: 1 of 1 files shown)


slimsag:
Thanks for sending this, and sorry I haven't had a chance to give it my full attention yet.

I'm concerned about the complexity of `doMountUnmount`, and the fact that it uses reflection. I think I need to consider more heavily in which situations users will need/want component mount and unmount events in the general case. One immediate question might be: Will normal users (not interacting with JS) want Mount to be a separate function than their component constructor `NewMyComponent`?

I'll delve into this further later. Sorry for the delays, again!

pdf:
I understand the lack of time, no worries.

The complexity in `doMountUnmount` is at least as much about correcting `Unmount` behaviour as it is about enabling mounting.  As it stands, `Unmount` is broken any time the number of siblings changes, which can lead to leaking handles and other unexpected behaviour in user code.  The added complexity to support `Mount` is actually pretty low, if you consider that most of the code is required to make `Unmount` work.

I really do hope you'll consider JS/DOM interaction a valid use-case (I suspect this may be a moderately regular hurdle), though I think it'd still be worthwhile simply for API consistency.  It also supports component re-use.

I'm currently building against my fork and using Mount/Unmount to manage JS/DOM interactions, as well as Store subscribe/unsubscribe.

slimsag:
> I really do hope you'll consider JS/DOM interaction a valid use-case (I suspect this may be a moderately regular hurdle), though I think it'd still be worthwhile simply for API consistency. It also supports component re-use.

FWIW I absolutely do consider that a valid use case. But I think these users will need very extensive integration, far more than what is provided here with simple Mount/Unmount. My theory is that we could allow these users to integrate via a `js` API directly, and reserve the Go Vecty API here for users who are not attempting JS/DOM interaction directly.

pdf:
I've updated this to treat component inequality as a mount/unmount trigger, rather than type equality as otherwise references to a component that leak to the outside world from the mount hook can lead to unexpected behaviour if the component is replaced with a new instance.  Combined with #94 though, additional mount/unmount events can be avoided by using a single reference to a component across renders.

pdf:

Issue #92, opened 2017-03-03
"Uncaught TypeError: Cannot read property 'style' of null" panic when Restorer interface implemented.


shurcooL:
I'm new to vecty, and I'm in the process of learning it and becoming more familiar with its features. I may be doing something illegal without realizing it, or this could be a bug in vecty implementation.

My implementation of `Restorer` looks like this:

```Go

type PresentationChanges struct {
	vecty.Core
	*model.RepoPresentation // Only uses Changes, and if len(Changes) == 0, then LocalRevision and RemoteRevision.
}

// Restore is called when the component should restore itself against a
// previous instance of a component. The previous component may be nil or
// of a different type than this Restorer itself, thus a type assertion
// should be used.
//
// If skip = true is returned, restoration of this component's body is
// skipped. That is, the component is not rerendered. If the component can
// prove when Restore is called that the HTML rendered by Component.Render
// would not change, true should be returned.
func (p *PresentationChanges) Restore(prev vecty.Component) (skip bool) {
	old, ok := prev.(*PresentationChanges)
	if !ok {
		return false
	}
	return old.RepoPresentation == p.RepoPresentation
}
```

The panic looks like this:

![image](https://cloud.githubusercontent.com/assets/1924134/23536723/e4914fd6-ff94-11e6-9e60-2a6f861cabaf.png)

It happens on this line:

```Go
style := h.Node.Get("style")
```

Inside `func (h *HTML) restoreHTML(prev *HTML)`, because `prev.Node` is `nil`, which gets assigned to `h.Node` on first line of `restoreHTML`.

### Reproduce

No smaller repro for now, but it should be very easy to repro with a large project. Here are the steps:

1. `go get -u -d github.com/shurcooL/Go-Package-Store/cmd/Go-Package-Store`.
2. `cd` into the repository and checkout latest version of `frontend-vecty-issue-92` branch.
3. Modify hardcoded path in line 19 of `./cmd/Go-Package-Store/updates.go` file to one that matches your local setup (sorry, it's a temporary development artifact).
4. `go build -tags=dev github.com/shurcooL/Go-Package-Store/cmd/Go-Package-Store`
5. Execute `Go-Package-Store` binary.
6. Open http://localhost:7043/updates in a browser.
7. You'll see [this](https://cloud.githubusercontent.com/assets/1924134/23536870/0a00ce62-ff96-11e6-89bd-7a9ad2ec62e9.png).
8. Click top-most "Update" button (not Update All).
[...]

Issue #78, opened 2017-01-26
Panic on nextChildRender == prevChildRender


pdf:
Can you tell me why https://github.com/gopherjs/vecty/blob/34ae77073fd923c6ab2f89c234a8ecf76f463cb0/dom.go#L175 exists?

Would it not be safe to just continue here?

shurcooL:
The commit message of ee9d6c9886074534c4ad3aadbc3c2df27247b284 says:

> panic if nextChildRender == prevChildRender (very invalid state)
>
> See #70

pdf:
Sorry, should have used blame to start, however I've read the issue and the related code, and I still may be confused.

The upshot seems to be that for every render, nested components must be reinstantiated.  I think I can see why that is as the code stands, but it makes natural patterns produce this panic, which is confusing.

The obvious pattern is mentioned in #70 - passing and storing components to be nested.  This can be worked around as Dave did by passing a closure that returns new instances of the components, but this doesn't seem ideal.

I think we can support at least Components but just calling Rerender if `prevChild == nextChild`, ie:

```diff
diff --git a/dom.go b/dom.go
index 7ee1c9c..12fa374 100644
--- a/dom.go
+++ b/dom.go
@@ -158,8 +158,8 @@ func (h *HTML) restoreHTML(prev *HTML) {

        // TODO better list element reuse
        for i, nextChild := range h.children {
-               nextChildRender := doRender(nextChild)
                if i >= len(prev.children) {
+                       nextChildRender := doRender(nextChild)
                        if doRestore(nil, nextChild, nil, nextChildRender) {
                                continue
                        }
@@ -167,6 +167,11 @@ func (h *HTML) restoreHTML(prev *HTML) {
                        continue
                }
                prevChild := prev.children[i]
+               if c, isComponent := nextChild.(Component); isComponent && prevChild == nextChild {
+                       Rerender(c)
+                       continue
+               }
+               nextChildRender := doRender(nextChild)
                prevChildRender, ok := prevChild.(*HTML)
                if !ok {
                        prevChildRender = prevChild.(Component).Context().prevRender
```

Thoughts?

slimsag:
Removing this panic is NOT a valid thing to do. While you may find a way to make this work technically in the code -- it's not a good thing to be doing in general: a component Render should _always_ propagate downwards into child components, rendering the entire tree of elements _from scratch_.

[...]

PR #94, opened 2017-03-14
Allow component re-use

pdf:
Partial fix for #78, only allows re-use of `*vecty.Component`, not
`*vecty.HTML`.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -158,8 +158,8 @@ func (h *HTML) restoreHTML(prev *HTML) {
 
 	// TODO better list element reuse
 	for i, nextChild := range h.children {
-		nextChildRender := doRender(nextChild)
 		if i >= len(prev.children) {
+			nextChildRender := doRender(nextChild)
 			if doRestore(nil, nextChild, nil, nextChildRender) {
 				continue
 			}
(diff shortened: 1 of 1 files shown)


slimsag:
Sorry I haven't had a chance to give this my full attention yet. I hope to investigate this on Sat/Sun.

Overall it looks good, but I'd like to make sure there aren't any regressions, which is tough to do. I'd like to land some unit tests for this code (I've been working on those, but haven't pushed anything yet)

pdf:
I've pushed alternative logic for handling this, rather than jumping out of the standard loop into `Rerender`, it just adds conditions to the panic.

dom.go
```diff
@@ -207,7 +207,7 @@ func (h *HTML) Restore(old ComponentOrHTML) {
 		}
 	}
 
-	if prev, ok := old.(*HTML); ok && prev != nil {
```
pdf: I suspect this may fix #92 as well

dom.go
```diff
@@ -207,7 +207,7 @@ func (h *HTML) Restore(old ComponentOrHTML) {
...
+	if prev, ok := old.(*HTML); ok && prev != nil && prev.Node != nil {
```
pdf: I suspect this may fix #92 as well

pdf:
I've had to revert to the original implementation, as the alternative was not reliably re-rending some children.

pdf:

PR #96, opened 2017-03-28
Disallow Property("style", ...).

### LINKED ISSUE #90: Document that using vecty.Style is preferable over vecty.Property("style", ...).
I've asked what's the difference between the two in https://gophers.slack.com/archives/vecty/p1488148154608710, and @slimsag gave a great answer/explanation.

It should be a part of vecty documentation, whether the godoc, or somewhere else that goes over these things.

https://godoc.org/github.com/gopherjs/vecty#Property does not say anything about "style" being a special case right now.

It does say:

> Generally, this function is not used directly but rather the style subpackage (which is type safe) is used instead.

But that seems a little unclear and not completely relevant.

shurcooL:
Update documentation to point out that [`style`](https://godoc.org/github.com/gopherjs/vecty/style) package or [`Style`](https://godoc.org/github.com/gopherjs/vecty#Style) should be used instead. Panic in `Property` if passed key "style".

Also update documentation of `Property` to point out both `prop` and `style` packages that are preferable. This was mentioned in https://github.com/gopherjs/vecty/issues/90#issuecomment-282601552.

Fixes #90.

This is what it looks like if the user breaks the API contract:

image

### DIFF
--- a/markup.go
+++ b/markup.go
@@ -87,7 +87,7 @@ func (m markupFunc) Apply(h *HTML) { m(h) }
 
 // Style returns Markup which applies the given CSS style. Generally, this
 // function is not used directly but rather the style subpackage (which is type
-// safe) is used instead.
+// safe) should be used instead.
 func Style(key, value string) Markup {
 	return markupFunc(func(h *HTML) {
 		if h.styles == nil {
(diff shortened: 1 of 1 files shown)


slimsag:
❤️ 

Thanks for this!

PR #99, opened 2017-03-28
Add .travis.yml for Travis CI.

### LINKED ISSUE #95: Add CI.
I found it surprising to see this Go project does not have any sort of CI.

I ran into an issue where some of the code wasn't gofmted. I fixed that and wanted to add a gofmt formatting check to `.travis.yml` or equivalent, but couldn't, because there's no CI.

CI should be added so such basic problems don't have to be detected manually.

shurcooL:
Don't try to build on `master`, because this package is expected to use GopherJS compiler, which only supports current version of Go.

Someone with admin access to this repository will still need to enable it at [travis-ci.org](https://travis-ci.org), then this `.travis.yml` configuration file will take effect. /cc @slimsag

Fixes #95.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+sudo: false
+language: go
+go:
+  - 1.8.x
+install:
+  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
+script:
+  - go get -t -v ./...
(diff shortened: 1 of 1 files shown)


slimsag:
> Someone with admin access to this repository will still need to enable it at travis-ci.org, then this .travis.yml configuration file will take effect. /cc @slimsag

Done.

Thanks for this!

PR #103, opened 2017-03-29
tests: travis: cache gopath

xoviat:


shurcooL:
Can you reference some documentation about how this works?

shurcooL:
Thanks for the explanation @xoviat. I'll use this as reference and look into it more later, when I find some time. Travis is passing successfully on master for now, so we can delay dealing with this for now.

Issue #31, opened 2016-02-13
Documentation


slimsag:

gedw99:
:) Even just the briefest would be cool..

anxiousmodernman:
I've sketched a very basic "hello world" tutorial here: https://github.com/gloriousfutureio/vecty/blob/31_docs/docs/TUTORIAL.md

The tutorial should focus on Vecty concepts and be short
* no external css/js
* demonstrate Component declaration, reuse 
* demonstrate how components are just structs, use custom fields

I'm a fan of go-kit's approach with their [stringsvc tutorial](https://gokit.io/examples/stringsvc.html#stringsvc1). We can start smaller than that.

anxiousmodernman:
@slimsag Has an outline for docs contribution on the **docfest** branch https://github.com/gopherjs/vecty/blob/docfest/docs/index.md#vecty-introduction

PR #105, opened 2017-04-02
Documentation and Tutorial updates

anxiousmodernman:
Add updates to README.md, and a simple hello world example.

Also add a docs folder, where we can collect long-form documentation. This will be
subject to change as the API evolves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,17 @@ Features
 -	XSS protection: unsafe HTML must be explicitly denoted as such.
 -	Reusability: share components by making Go packages that others can import!
 
+Examples
+========
+
+Run examples with the gopherjs development server
+
+```
+gopherjs serve
+```
+
+Then navigate to `http://localhost:8080/github.com/gopherjs/vecty/examples`. 
+
 Goals
 =====
 
--- a/docs/COMPONENTS.md
+++ b/docs/COMPONENTS.md
@@ -1,0 +1,92 @@
+# Components
+
+There are two ways to define something that will render on a page.
+
+1. Create a `struct` that embeds `vecty.Core` and implements `Render`
+2. Write a function with the signature `func(markup ...vecty.MarkupOrComponentOrHTML) *vecty.HTML`
+
+## Component Structs
+
+Component structs look like this
+
+```go
+type SomeComponent struct {
+    vecty.Core
+}
+
+func (sc *SomeComponent) Render() *vecty.HTML {
+    return elem.Body(vecty.Text("some text in the body"))
+}
+```
+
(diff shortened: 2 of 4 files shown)


anxiousmodernman:
I intend to squash these down after discussion or approval. I think it would be helpful to have something a little bit "official" around contributions and project structure. We might want to adopt/link to something from gopherjs.

Having contribution guidelines documented will make it easier for maintainers. I have a boilerplate one here https://gist.github.com/anxiousmodernman/0ef4bb8441a3771a57bd3cc7b5b497b4 but since this is being developed under gopherjs we might want @shurcooL input.

slimsag:
Hi @anxiousmodernman ! I won't have time to review this in-depth for at least 1 week, apologies ahead of time. A couple of high level comments right off the bat:

- Any PR for docs should be made against the [`docfest` branch](https://github.com/gopherjs/vecty/tree/docfest), please. The reason this is not being done on `master` is because documentation cannot be fully written until at least 1.0 (https://github.com/gopherjs/vecty/milestone/1) which is some ways off. Additionally, I would like to keep docs hidden from the general public for the time being until I do feel Vecty is in a 1.0 state / ready for broad adoption.

I'll have further comments later, I am sure, but in general I would like to say that documentation is the one thing I think Vecty must absolutely do right. This means it needs to be reviewed and revised extensively, which makes PRs a hard format to work in, I think, due to the latency between me and you. Regardless, as long as that doesn't bother you I'm glad to have such contributions :) For the sake of making such PRs less long-winded, could we edit only one documentation file per PR? I think that would help.

slimsag:
>  I think it would be helpful to have something a little bit "official" around contributions and project structure.

Can you elaborate on what, specifically, you would've liked to find in such a document? My personal preference lies towards such documents being a waste of time / resources because they typically only cover things that are obvious; and for things that are not so obvious the document typically can't describe them in an easy way / they have to be described via direct communication anyway.

anxiousmodernman:
I will close this and reopen against `docfest` later. I see in that branch you also have a **docs** folder  with markdown inside.

Regarding the second point, many projects have a **CONTRIBUTING.md** file in their root. This is a loose convention that [GitHub endorses with some light tooling](https://github.com/blog/1184-contributing-guidelines). Examples of things you can do there:

* Guidelines on naming git branches
* Commit style, squash vs no-squash
* Code Style guides, or a link to a common one like Golange Code Comments
* Outline of project structure (where code lives, how to build, how to run tests, where docs live)
* How to file bugs

People will usually end up reaching out directly. But if you, as the maintainer, need to take a hard line to protect the quality of the project, having a doc to reference can be helpful.

I'll close this for now in favor of discussion on slack or in issue #31

Issue #106, opened 2017-05-07
Restorer is not supported by Rerender


pdf:
As `prev` is always passed as `nil` from `Rerender` to `doRestore`, there's no way for Components to function as a Restorer on Rerender.

https://github.com/gopherjs/vecty/blob/34ae77073fd923c6ab2f89c234a8ecf76f463cb0/dom.go#L298

Considering this is probably the primary use-case, it renders Restorer currently useless, except where used internally.

As it's currently broken for all(?) client code, if this can be fixed I'd also like to propose breaking the API to rename the method to something that parses lexically, like `ShouldUpdate` so that we can flip the boolean meaning.

It currently parses very painfully, since you implement a function called `Restore`, but what it actually means means is "don't update the DOM" if the return value is `true`.  Even if if it was readily apparent what `Restore` means from outside the rendering code, `Restore() -> true` meaning "don't restore" is a headache.

pdf:
@slimsag I'm going to take a stab at fixing this, and in the process take a pass over the whole render chain, including tackling #25 and #92.  This may be invasive, but the current render chain seems overly brittle.

I realise that you're keen to get some tests in before accepting PRs, and I understand that you've had little time to work on this project recently, however I'd like to get something more robust working than what we currently have so that I can move on with my project.

My current plan is to build on #93 and #104, and sort out #94 using a different approach during the refactor.

If you have a moment to comment, and would prefer that I don't do any of this, or would suggest that this would be wasted, I'll put this aside and possibly take a look at some of the React bindings that seem to be maturing for this project.  I'd really like to see Vecty succeed though, would like to stay native if possible, and am happy to invest some time if warranted, but I need to commit to a path to move forward.

pdf:
I went ahead and did it anyway, as I got into the flow last night: #107.

slimsag:
For better context, I responded to that comment inline here instead: https://github.com/gopherjs/vecty/pull/107#issuecomment-300897887

Issue #84, opened 2017-02-04
Integrating with js Libraries


newlix:
Can I use JS libaries (ex. http://www.dropzonejs.com/#create-dropzones-programmatically)  with vecty?  What's the idiomatic way?

slimsag:
Hi @newlix -- I unfortunately don't have time to dig into the specifics of that library right now, but I assume you're referencing this example:

```
// Dropzone class:
var myDropzone = new Dropzone("div#myId", { url: "/file/post"});
```

i.e., you're wondering what you would pass in as `div#myId` in the case of an element rendered by Vecty. Given that is the case, the way you would do that would be to assign an `Id` to your element in Vecty. e.g.:

```
func (m *MyComponent) Render() *vecty.HTML {
	return elem.Div(
                prop.ID("myId"),
	)
}
```

Unfortunately, there isn't yet a way for you to be informed via Vecty that "your element has been created" for you to actually instantiate `Dropzone` in the example code above. This is due to issue https://github.com/gopherjs/vecty/issues/75

A workaround until that issue is fixed would be something like e.g. querying the DOM constantly until that element appeared, then instantiating `Dropzone`.

newlix:
Hi, thanks for your suggestion.

I did solve the problem.   I was wondering why I couldn't get the dom element after vecty.RenderBody was called. After reading the RenderBody source code, I found that body would only be rendered after the doc is ready.
```
// RenderBody renders the given component as the document body. The given
// Component's Render method must return a "body" element.
func RenderBody(body Component) {
	nextRender := doRender(body)
	if nextRender.tag != "body" {
		panic(fmt.Sprintf("vecty: RenderBody expected Component.Render to return a body tag, found %q", nextRender.tag))
	}
	doRestore(nil, body, nil, nextRender)
	// TODO: doRestore skip == true here probably implies a user code bug
	doc := js.Global.Get("document")
	if doc.Get("readyState").String() == "loading" {
		doc.Call("addEventListener", "DOMContentLoaded", func() { // avoid duplicate body
			doc.Set("body", nextRender.Node)
		})
		return
	}
	doc.Set("body", nextRender.Node)
} 
```

I found a easier way to make sure that the client.js generated by gopherjs will not run until after the page has loaded by using the HTML <script> defer Attribute (http://www.w3schools.com/TAgs/att_script_defer.asp).
```
//index.html
...
<script src="client.js" defer></script>
```
[...]

Issue #25, opened 2016-02-13
Better list element reuse / "keys"


slimsag:
See https://github.com/gopherjs/vecty/blob/master/dom.go#L114

shurcooL:
@slimsag, you should post permalinks to lines in files. The `Y` shortcut is extremely helpful for making them. Using `master` branch means a line number is likely to point to something else after time passes and things shift around.

I went back to Feb 13th when the issue was created and found the commit you must've been referring to:

https://github.com/gopherjs/vecty/blob/8bd83c54933703ee2a802c6fcd39366b855a9e22/dom.go#L114

It pointed to a line containing `// TODO better list element reuse` comment. I can see that comment is still available in today's `master` some lines below:

https://github.com/gopherjs/vecty/blob/34ae77073fd923c6ab2f89c234a8ecf76f463cb0/dom.go#L159

(Feel free to edit your original issue and delete this comment to reduce noise, if you wish.)

PR #107, opened 2017-05-10
Render new

pdf:
Refactor rendering for reliability and performance.

NEW:
- `Keyer` interface provides support for efficient list element re-use.
  Components simply need to implement the `Key` function, or the
  `vecty.Key` Markup may be used for HTML.

BREAKING:
- `Restorer` interface is now `Updater`, and its `Restore` method is now
  `ShouldUpdate`. The previous `Restorer` implementation was variously
  broken in any case, and poorly named IMO.

  Additionally, it requires a `Clone` method to support passing a
  previous Component instance to `ShouldUpdate`.  There is a limited
  generic version available on `vecty.Core` using reflection that may
  satisfy for users that do not wish to implement it themselves.
- `Context` method on `Component` interface is no longer public.
- `Restore` method on HTML is no longer public.

Supersedes: #93 #94 #104 
Impacts: #84 #75 #54
Fixes: #106 #92 #85 #78 #25


dom.go
```diff
@@ -2,18 +2,44 @@ package vecty
...
+	"reflect"
 
 	"github.com/gopherjs/gopherjs/js"
 )
 
-// Core implements the Context method of the Component interface, and is the
-// core/central struct which all Component implementations should embed.
+var (
+	errMissingParent = fmt.Errorf("Missing parent node")
+	errEmptyElement  = fmt.Errorf("Empty element or node")
```
shurcooL: Minor Go style issue, error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.

Source: https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

dom.go
```diff
@@ -58,45 +91,248 @@ type Unmounter interface {
...
+	// be slower, and will panic if the Component is not a pointer to struct.
+	//
+	// The `current` argument is guaranteed to be equal to the Component, and so
+	// may be ignored by implementers - it is only there to serve the generic
+	// implementation in vecty.Core.
+	Clone(current Component) Component
+}
+
+// Keyer is an optional interface that Components may implement to uniquely
+// identify an instance amongst a list of its siblings.  Implementing Keyer
```
shurcooL: Another minor style issue. Go style uses a single space between sentences.

Source: https://dmitri.shuralyov.com/idiomatic-go#single-spaces-between-sentences.

dom.go
```diff
@@ -2,18 +2,44 @@ package vecty
...
+	"reflect"
 
 	"github.com/gopherjs/gopherjs/js"
 )
 
-// Core implements the Context method of the Component interface, and is the
-// core/central struct which all Component implementations should embed.
+var (
+	errMissingParent = fmt.Errorf("missing parent node")
+	errEmptyElement  = fmt.Errorf("empty element or node")
```
shurcooL: Also, minor, since these are static errors and you're not using any %v verbs, you could use the simpler `errors.New` instead of `fmt.Errorf`.

pdf:
We may be able to make an additional improvement by reducing reflows using a single append to the DOM via a DocumentFragment, but I need to work out how best to handle deferring lifecycle (mount/unmount) events there, since they need to be triggered for children after the DOM change.  Also need to determine which cases this is optimal for, but should be a safe bet for new elements at a minimum.

slimsag:
Thanks for sending this. I'm on a vacation, so I now have more spare time. I should have time to merge these changes soon. I'll respond to your other comment https://github.com/gopherjs/vecty/issues/106#issuecomment-300138104 here for better context.

> I realise that you're keen to get some tests in before accepting PRs, and I understand that you've had little time to work on this project recently, however I'd like to get something more robust working than what we currently have so that I can move on with my project.

I totally understand your desire to push this forward. I'll be doing my best very soon (hopefully today) to take a look at these changes one-by-one and add tests for them while doing so. I should point out that it is hard to work on such a large PR as this, so I'll be personally breaking up your changes into smaller segments and referencing this PR. I'll keep your git author data attached, though, so it doesn't look like I am stealing your work :P I hope that's alright!

Although I would certainly have preferred smaller more scoped PRs, I just wanted to say thanks again for doing this! Your work was not a wasted effort, and I appreciate you taking the time to contribute!

Will be back with more soon.

pdf:
@slimsag write:
>  I should point out that it is hard to work on such a large PR as this

Yes, and I apologise, but with the scope of the changes I wanted to make I couldn't see much in the way of alternatives - I essentially rewrote the render chain, and not much more than the prop/attr/etc assignment code remains of the original.

I tried to include a decent level of commenting, but if there are any explanations or clarifications I can provide that will ease your efforts, please don't hesitate to ask.

pdf:
Rebased against master with the new node structure.

slimsag:
Thanks for the rebase. I'm still working on tests against the master branch right now. Once that's done I will be picking apart portions of your branch to review and merge separately. I'll handle rebasing your branch in this repo once I begin doing that

pdf:
@slimsag FYI, I have one bug to fix here that I'm aware of, related to a scenario where Rerender is called on a Component that has already been replaced.

pdf:
The [above-mentioned bug](https://github.com/gopherjs/vecty/pull/107#issuecomment-301717454) should be fixed by d0d5cd0, which stops components from being re-rendered directly after they're removed, but retains the ability to pass components around and render them as children.

The latest commit (242ab03) adds batching via `requestAnimationFrame`, and deduplication of Component re-renders.  I believe deduplication in this fashion should be safe since any component implementation I can think of must manage state internally, so we're just skipping over intermediate states.  If you believe that there are some conditions where this may not be safe, we can omit dedup, but it's working well here.  I may take one more pass over the code, but I wanted to get it out to obtain some feedback.

The raf dependency/polyfill I've pulled in is pretty small, so we could just add it locally if you'd prefer not to carry an external dependency for it.

slimsag:
https://github.com/gopherjs/vecty/pull/107/commits/ecbef2a69872f498e058ff6e744856bdeef7032c, "Add support for namespaced elements", is the first part of your change that I've taken a look at. I've factored it out into https://github.com/gopherjs/vecty/pull/110

slimsag:
-> https://github.com/gopherjs/vecty/pull/112 for rebased branch

pdf:
I think we may as well close this for now, let's work on getting changes in one at a time.

Issue #29, opened 2016-02-13
determine how best to perform testing


slimsag:
as it requires a browser DOM to be present, etc.

influx6:
I would definitely like to see the testing approach that you guys will take, currently once can only test non DOM based code. Would be nice to get such level of tests for the visual end and dom based functions as well.

neelance:
I'd like to have something like React's shallow rendering, see https://facebook.github.io/react/docs/test-utils.html#shallow-rendering. Such testing is easy to reason about, tests only one component at a time, requires no DOM and thus might even run with `gc`.

slimsag:
That sounds great @neelance ! I like the idea of that a lot :) I will see if I can find time to look into it.

joeblew99:
I also think that component based testing is the way to initially go. They are unit tests of that unit / component.

pdf:
@slimsag I'm struggling a little bit trying to determine the best way to test more complex element hierarchies.  I feel like maybe we need a higher-order wrapper for the `mockObject`, but I'm not entirely sure of the best way to track element identity, to test for correct order, etc - ideally I'd like to sort this out in a way that allows table-driven testing for maintenance.

pdf:
I'm pretty much blocked on doing any further work until we either decide what to do about #115, or work out how to go about testing trees of elements.

slimsag:
Sorry I haven't had a chance to look at this yet @pdf -- I think we just need to add some more extensive helper utility functions to our test suite but I haven't had time to figure out which are needed / write them. I'll gladly accept idea's/PRs in this area if anyone wants to beat me to it.

I've responded on #115 and think we can find a solution there very soon (perhaps today)

PR #108, opened 2017-05-11
enable mocking JS calls for testing Vecty

slimsag:
**Breaking:** `(*HTML).Node` is now a function instead of a struct field.

This change enables us to mock JS calls and ensure that Vecty performs the right
operations internally when something occurs. Actual tests will be added in
a follow-up PR, and will be compatible with `go test` (instead of `gopherjs test`).

Additionally, a `docs/CHANGELOG.md` file is added to communicate breaking changes clearly to users.

Helps https://github.com/gopherjs/vecty/issues/29 and https://github.com/gopherjs/vecty/pull/107#issuecomment-300897887

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -77,7 +77,7 @@ type Restorer interface {
 // HTML represents some form of HTML: an element with a specific tag, or some
 // literal text (a TextNode).
 type HTML struct {
-	Node *js.Object
+	node jsObject
 
 	tag, text, innerHTML   string
 	styles, dataset        map[string]string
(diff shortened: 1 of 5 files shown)


dom.go
```diff
@@ -346,25 +348,76 @@ func RenderBody(body Component) {
...
+	String() string
+	Bool() bool
+}
+
+func wrapObject(j *js.Object) jsObject {
+	if j == nil {
+		return nil
+	}
+	if j == js.Undefined {
+		panic("TODO")
```
pdf: Can we not just return `js.Undefined`?  I tried rebasing my PR against master, but this breaks things pretty badly.

slimsag: Let's add that as a new commit to your PR :) I don't want to add anything that is not needed here

PR #110, opened 2017-05-17
Add Namespace for creating namespaced elements (e.g. for inline SVG)

slimsag:
This allows the creation of elements like inline SVG.

Helps #54

Author amended to @pdf instead of me (@slimsag), to provide credit for his
original change at https://github.com/gopherjs/vecty/pull/107

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -79,11 +79,11 @@ type Restorer interface {
 type HTML struct {
 	node jsObject
 
-	tag, text, innerHTML   string
-	styles, dataset        map[string]string
-	properties, attributes map[string]interface{}
-	eventListeners         []*EventListener
-	children               []ComponentOrHTML
+	namespace, tag, text, innerHTML string
+	styles, dataset                 map[string]string
+	properties, attributes          map[string]interface{}
+	eventListeners                  []*EventListener
+	children                        []ComponentOrHTML
 }
 
 func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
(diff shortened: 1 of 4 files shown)


slimsag:
I've factored this change out of https://github.com/gopherjs/vecty/pull/107 (https://github.com/gopherjs/vecty/commit/ecbef2a69872f498e058ff6e744856bdeef7032c) and made some changes to it. Below is my response to the original change:

On first glance, I don't like `NamespacedTag` being different than `Tag` when, at least in the DOM API, the namespace is really just associated with the tag (it's not strictly a different type of tag, per-say). So I'd prefer `Tag("a", Namespace("b"))` to handle this situation instead. This matches `UnsafeHTML` nicely.

I also found [`setAttributeNS`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttributeNS) referenced quite extensively online, but later found through lack of references to it in the SVG2 spec that it seems to not be needed (which is good / means we won't need a `NamespacedAttribute` variant too).

I also noticed that you made `Tag` an alias for `NamespacedTag` with namespace `""`, which seems less correct to me when creating HTML elements, although I could not find any strict definition of which is more correct. The MDN docs make only vague references to valid namespace URIs without listing an empty string as one (https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS#Valid Namespace URIs), and does state:

> To create an element without specifying a namespace URI, use the createElement method.

I also wrote some alternative docs that do not reference the `elem` subpackage in the `NamespacedTag` function (now `Namespace`) as that is very incorrect for that function, and altered `Attribute` docs to be non-HTML specific (as it works with SVG, too).

slimsag:
I'll merge this change now, but am open to discussing all aspects of this change (let's use the issue https://github.com/gopherjs/vecty/issues/54 for that discussion, or on a new PR somewhere).

pdf:
Something that was missing from this change, but is superseded by my other changes:

https://github.com/gopherjs/vecty/blob/d97ce692ac1f6bcdf5f88321f0bcc3c002c9ec95/dom.go#L217

Probably needs to check namespace equality too, before restoration.  Probably not a likely occurrence that a tag matches across namespaces, but needed for correctness.

slimsag:
Excellent point; I missed that as it's currently down the untested code path. I've spotted a second issue there as well, around TextNode restoration.

Sent https://github.com/gopherjs/vecty/pull/111 for this.

PR #114, opened 2017-05-18
Rerender improvements

### LINKED ISSUE #106: Restorer is not supported by Rerender

slimsag:
- Invoking `Rerender` on a component with no previous render now panics with a more helpful error message.
- General `Rerender` documentation improvements.
- Improve panic message when a user invokes `Rerender(nil)` by accident.
- Fix bug with `Rerender` not invoking the component `Restore` method with the previous component.

Fixes #106 
Helps #92

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -10,8 +10,8 @@ import (
 // Core implements the Context method of the Component interface, and is the
 // core/central struct which all Component implementations should embed.
 type Core struct {
-	prevComponent Component
-	prevRender    *HTML
+	prevComponent, prevRenderComponent Component
+	prevRender                         *HTML
 }
 
 // Context implements the Component interface.
@@ -70,19 +70,34 @@ type Unmounter interface {
 type ComponentOrHTML interface{}
 
 // Restorer is an optional interface that Component's can implement in order to
-// restore state during component reconciliation and also to short-circuit
-// the reconciliation of a Component's body.
+// restore state during component reconciliation.
 type Restorer interface {
 	// Restore is called when the component should restore itself against a
-	// previous instance of a component. The previous component may be nil or
-	// of a different type than this Restorer itself, thus a type assertion
-	// should be used.
+	// previous instance of a component.
 	//
-	// If skip = true is returned, restoration of this component's body is
-	// skipped. That is, the component is not rerendered. If the component can
-	// prove when Restore is called that the HTML rendered by Component.Render
-	// would not change, true should be returned.
-	Restore(prev Component) (skip bool)
+	// The previous component may be nil or of a different type than this
+	// Restorer itself, thus a type assertion should be used no action taken
+	// if the type does not match.
+	Restore(prev Component)
+}
+
+// RenderSkipper is an optional interface that Component's can implement in
+// order to short-circuit the reconciliation of a Component's rendered body.
+//
+// This is purely an optimization, and does not need to be implemented by
+// Components for correctness. Without implementing this interface, only the
+// difference between renders will be applied to the browser DOM. This
+// interface allows components to bypass calculating the difference altogether
+// and quickly state "nothing has changed, do not re-render".
+type RenderSkipper interface {
+	// SkipRender is called with a copy of the Component made the last time its
+	// Render method was invoked. If it returns true, rendering of the
+	// component will be skipped.
+	//
+	// The previous component may be of a different type than this
+	// RenderSkipper itself, thus a type assertion should be used no action
+	// taken if the type does not match.
+	SkipRender(prev Component) bool
 }
 
 // HTML represents some form of HTML: an element with a specific tag, or some
@@ -367,18 +382,28 @@ func doRenderHTML(c ComponentOrHTML, prevRender *HTML) (h *HTML, skip bool) {
 }
 
 // doRender handles rendering the given Component into *HTML. If skip == true
-// is returned, the Component's Restore method has signaled the component does
+// is returned, the Component's SkipRender method has signaled the component does
 // not need to be rendered and h == nil is returned.
 func doRender(comp Component) (h *HTML, skip bool) {
-	// Before rendering, consult the Component's Restore method to see if we
+	// Before rendering, consult the Component's SkipRender method to see if we
 	// should skip rendering or not.
-	if r, ok := comp.(Restorer); ok {
-		prev := comp.Context().prevComponent
-		if comp == prev {
-			panic("vecty: internal error (Restore called with identical prev component)")
+	if rs, ok := comp.(RenderSkipper); ok {
+		prev := comp.Context().prevRenderComponent
+		if prev != nil {
+			if comp == prev {
+				panic("vecty: internal error (Restore called with identical prev component)")
+			}
+			if rs.SkipRender(prev) {
+				return nil, true
+			}
 		}
-		if skip := r.Restore(prev); skip {
-			return nil, true
+	}
+
+	// Now that we know we are rendering the component, restore friendly
+	// relations between the component and the previous component.
+	if comp != comp.Context().prevComponent {
+		if r, ok := comp.(Restorer); ok {
+			r.Restore(comp.Context().prevComponent)
 		}
 	}
 
@@ -394,7 +419,8 @@ func doRender(comp Component) (h *HTML, skip bool) {
 
 	// Update the context to consider this render.
 	comp.Context().prevRender = nextRender
-	comp.Context().prevComponent = doCopy(comp)
+	comp.Context().prevComponent = comp
+	comp.Context().prevRenderComponent = doCopy(comp)
 	return nextRender, false
 }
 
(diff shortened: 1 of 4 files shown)


pdf:
Sorry for the delay getting back to this - my only concern here from a usability viewpoint is that it will be somewhat difficult for users to reason about what expected results are if they instantiate a component with fields populated, that also has a Restore method.

In the following scenario:
```go
type OuterComponent struct {
        vecty.Core
        OuterField bool
}

func (c *OuterComponent) Render() *vecty.HTML {
        return elem.Div(
                &InnerComponent{
                        InnerField: c.OuterField,
                },
        )
}

type InnerComponent struct {
        vecty.Core
        InnerField bool
}

func (c *InnerComponent) Restore(old vecty.Component) {
        oldInner, ok := old.(*InnerComponent)
        if !ok {
                return
        }

        c.InnerField = oldInner.InnerField
}

func (c *InnerComponent) Render() *vecty.HTML {
        return elem.Div()
}
```

If `OuterComponent.OuterField` changes between renders, users may expect `InnerComponent.InnerField` to equal the new `OuterField` value, but instead it will equal the value from the previous instance of `InnerField`.

I'm not sure if we can or should do anything to help resolve this, but if we don't, it will need to be very clearly documented that fields `Restore`d should not overlap with fields set from the parent.

dom.go
```diff
@@ -297,61 +320,120 @@ func Text(text string, m ...MarkupOrComponentOrHTML) *HTML {
...
+// not need to be rendered and h == nil is returned.
+func doRender(comp Component) (h *HTML, skip bool) {
+	// Before rendering, consult the Component's SkipRender method to see if we
+	// should skip rendering or not.
+	if rs, ok := comp.(RenderSkipper); ok {
+		prev := comp.Context().prevRenderComponent
 		if prev != nil {
-			p = prev.(Component)
+			if comp == prev {
+				panic("vecty: internal error (Restore called with identical prev component)")
```
pdf: Restore != SkipRender

dom.go
```diff
@@ -297,61 +320,120 @@ func Text(text string, m ...MarkupOrComponentOrHTML) *HTML {
...
-	return r
 }
 
-func doRestore(prev, next ComponentOrHTML, prevRender, nextRender *HTML) (skip bool) {
-	if r, ok := next.(Restorer); ok {
-		var p Component
+// doRender handles rendering the given Component into *HTML. If skip == true
+// is returned, the Component's SkipRender method has signaled the component does
+// not need to be rendered and h == nil is returned.
+func doRender(comp Component) (h *HTML, skip bool) {
```
pdf: If Restore happened before SkipRender, wouldn't that allow the Component to determine whether there was a state change, allowing us to use prevRender (unless nil) from prevComponent if no update is necessary?

dom.go
```diff
@@ -297,61 +320,120 @@ func Text(text string, m ...MarkupOrComponentOrHTML) *HTML {
...
-func renderHandleNil(c Component) *HTML {
-	r := c.Render()
-	if r == nil {
-		// nil renders are translated into noscript tags.
-		r = Tag("noscript")
+// doRenderHTML renders the ComponentOrHTML. In the case of *HTML, its Restore
+// method is invoked with the specified prevRender and c.(*HTML) is returned.
+//
+// In the case of a Component, doRender(c.(Component)) is returned.
+func doRenderHTML(c ComponentOrHTML, prevRender *HTML) (h *HTML, skip bool) {
```
pdf: Naming is confusing because this renders either HTML or Component.

pdf:
Rerender is also not safe to process for Components that have already been removed from the DOM (d0d5cd0b9236925da4051bd766d6e6d5faef18a3), though this can be addressed later.

slimsag:
> my only concern here from a usability viewpoint is that it will be somewhat difficult for users to reason about what expected results are if they instantiate a component with fields populated, that also has a Restore method.

Hmm, I see your point. That's quite tricky, and a very subtle issue that I think would be bad for users to run into. Could you point me to how your branch handled this case? (did it?) I also wonder what happens in a similar situation with React. We should also add a test for this exact use case -- regardless of what is decided.

dom.go
slimsag: I agree that is better 👍

dom.go
slimsag: 👍

dom.go
slimsag: Good catch. Not sure why i didn't think of that when working on this. I agree it would be better to move the component `Restore` call above the `SkipRender` call here. Should also add a test to make sure that it is invoked in that order to prevent any future change from altering the order without considering that case.

pdf:
>  Could you point me to how your branch handled this case? (did it?)

My branch didn't handle restoration at all, as the use-case wasn't documented and it hadn't occurred to me.

> I also wonder what happens in a similar situation with React.

In React, there are two separate entities: `Props` which are attributes set from the parent scope; and `State` which is internal to the local Component.  The separation means this can't occur.

slimsag:
Got it. I took a look at https://facebook.github.io/react/docs/state-and-lifecycle.html though, and I'm left wondering what would prevent me from doing this in React:

```diff
class Clock extends React.Component {
  constructor(props) {
    super(props);
-    this.state = {date: new Date()};
+    this.state = {date: props.date};
  }

  render() {
    return (
      <div>
        <h1>Hello, world!</h1>
        <h2>It is {this.state.date.toLocaleTimeString()}.</h2>
      </div>
    );
  }
}
```

If the answer is that nothing really prevents that (aside from convention or perhaps a linter warning), then I suppose clearly documenting the difference would be enough. i.e. to say that anything copied over inside `Restore` (state) should not be set by a parent component externally (props). We could of course add a linter down the road to detect such issues, as well.

pdf:
> I'm left wondering what would prevent me from doing this in React

The constructor is only ever called once, so initializing state from props in the constructor is perfectly valid.  Once the Component is instantiated, while the Component exists in the (virtual) DOM all following updates to Props/State trigger lifecycle events rather than reinstantiating the whole Component.

pdf:
This quote from my https://github.com/gopherjs/vecty/pull/112#issuecomment-302653109 is relevant:

> I think in React this works because there's essentially transparent state persistence between Components of the same type at the same location, and a child is only ever updated if its Props are changed by the parent or its State changes internally, whereas Vecty requires all children to be re-rendered for every parent render, and if instantiated inline, two Components of the same type, at the same location, are separate entities that need to be manually migrated.
>
> The React model looks more like how Vecty behaves (with the modifications to allow it) if child components are persistent references (ie, a single component pointer is re-used between renders), since then the child component is able to maintain internal state between renders.

I wonder if we shouldn't consider some sort of method for registering children so that we can always re-use them between renders and remove the need for Restore completely.  Not sure how feasible elegent this might be though.

slimsag:
Oof. Okay, for some reason I really did think that the Component was instantiated on each render invocation. I was under the impression React silently copied `state` between component instances for the user. Given that, this is def. a problem that we'll need to address. I also see now that I didn't correctly understand the quoted comment above the first time I read it.

For now, let's not block this PR on fixing that (since AFAIK it will already be a problem today). I'm not 100% sure what the most elegant solution will look like... but I agree that maybe keeping child instances around somehow might be in order. (also, I responded to your email just now as well)

pdf:
Alright, we'll ignore this for now, but I think it would be worthwhile to try and work out how to handle this sooner rather than later, since this may have significant implications for the rest of the render pipeline.  I've created #115 to track this discussion.

slimsag:
I agree, thanks for opening that issue

pdf:
I think SkipRender must be omitted whenever prevRender is nil so that initial rendering can succeed.  This will make the panic in RenderBody impossible to trigger, so I'd probably drop that and replace the tests that depend on it with a local panic in the test code to short-circuit slightly earlier.  Sound correct to you?

slimsag:
> Scratch that - since users must type-assert in their SkipRender to do anything useful, and prev will be nil or incompatible on first component render, I think it's probably fine to leave this as-is.

This is exactly correct (the type assertion avoids the case of nil / handles first render).

slimsag:
Since all comments were addressed by your PR  #116 -- I'm going to merge this now. I think https://github.com/gopherjs/vecty/issues/115 is extremely important to consider / fix next, so I'll be thinking about that a bit & will post there if I come up with any ideas.

PR #117, opened 2017-06-16
HTML restore cleanup

pdf:
This probably is the largest change I'd like to land, so I thought I'd try to get this out up front.  I've split it up a little so that you can hopefully understand how I've arrived here (will squash before merge), however I'm hoping that the final changeset is not too large for comprehension.

The 10000ft view is that the majority of `*HTML.Restore` is redundant code, since `*HTML.restoreHTML` also performs these mutations.  Additionally, the name `restore` is a misnomer here, because the code is actually about performing DOM mutations, and the code has never implemented `Restorer`, even though the comment suggests otherwise.  I also can't see a reason for this code to be public.

Knowing whether we are mutating a compatible element from the previous render, or need to replace an incompatible element will allow us to make some more intelligent decisions about DOM manipulation in subsequent changes.

I will note that we currently appear to be missing tests for all the HTML mutations (e.g. properties, attributes, styles, etc), so I'd like to get them added either before this is merged, or as part of this PR.  One other thing to note is that this change highlighted that we were not previously mutating `dataset` between renders, only setting it on initial render.

Please let me know if this is comprehensible, though I'm not sure how I would make this one much smaller.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -115,6 +115,21 @@ type HTML struct {
 // Node returns the underlying JavaScript Element or TextNode.
 func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
 
+func (h *HTML) createNode() {
+	switch {
+	case h.tag != "" && h.text != "":
+		panic("vecty: only one of HTML.tag or HTML.text may be set")
+	case h.tag == "" && h.innerHTML != "":
+		panic("vecty: only HTML may have UnsafeHTML attribute")
+	case h.tag != "" && h.namespace == "":
+		h.node = global.Get("document").Call("createElement", h.tag)
+	case h.tag != "" && h.namespace != "":
+		h.node = global.Get("document").Call("createElementNS", h.namespace, h.tag)
+	default:
+		h.node = global.Get("document").Call("createTextNode", h.text)
+	}
+}
+
 func (h *HTML) restoreText(prev *HTML) {
 	h.node = prev.node
 
@@ -250,20 +265,8 @@ func (h *HTML) Restore(prev *HTML) {
 		}
 	}
 
-	if h.tag != "" && h.text != "" {
-		panic("vecty: only one of HTML.tag or HTML.text may be set")
-	}
-	if h.tag == "" && h.innerHTML != "" {
-		panic("vecty: only HTML may have UnsafeHTML attribute")
-	}
-	switch {
-	case h.tag != "" && h.namespace == "":
-		h.node = global.Get("document").Call("createElement", h.tag)
-	case h.tag != "" && h.namespace != "":
-		h.node = global.Get("document").Call("createElementNS", h.namespace, h.tag)
-	default:
-		h.node = global.Get("document").Call("createTextNode", h.text)
-	}
+	h.createNode()
+
 	if h.innerHTML != "" {
 		h.node.Set("innerHTML", h.innerHTML)
 	}
--- a/dom.go
+++ b/dom.go
@@ -130,7 +130,7 @@ func (h *HTML) createNode() {
 	}
 }
 
-func (h *HTML) restoreText(prev *HTML) {
+func (h *HTML) reconcileText(prev *HTML) {
 	h.node = prev.node
 
 	// Text modifications.
@@ -139,8 +139,37 @@ func (h *HTML) restoreText(prev *HTML) {
 	}
 }
 
-func (h *HTML) restoreHTML(prev *HTML) {
-	h.node = prev.node
+func (h *HTML) reconcile(prev *HTML) {
+	// Check for compatible tag and mutate previous instance on match, otherwise start fresh
+	switch {
+	case prev != nil && h.tag == "" && prev.tag == "":
+		// Compatible text node
+		h.reconcileText(prev)
+		return
+	case prev != nil && h.tag != "" && prev.tag != "" && h.tag == prev.tag && h.namespace == prev.namespace:
+		// Compatible element node
+		h.node = prev.node
+	default:
+		// Incompatible node, start fresh
+		if prev == nil {
+			prev = &HTML{}
+		}
+		h.createNode()
+	}
+
+	// Wrap event listeners
+	for _, l := range h.eventListeners {
+		l := l
+		l.wrapper = func(jsEvent *js.Object) {
+			if l.callPreventDefault {
+				jsEvent.Call("preventDefault")
+			}
+			if l.callStopPropagation {
+				jsEvent.Call("stopPropagation")
+			}
+			l.Listener(&Event{Object: jsEvent, Target: jsEvent.Get("target")})
+		}
+	}
 
 	// Properties
 	for name, value := range h.properties {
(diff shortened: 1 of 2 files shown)


dom.go
```diff
@@ -139,8 +139,37 @@ func (h *HTML) restoreText(prev *HTML) {
...
+		h.node = prev.node
+	default:
+		// Incompatible node, start fresh
+		if prev == nil {
+			prev = &HTML{}
+		}
+		h.newNode()
+	}
+
+	// Wrap event listeners
```
pdf: The following block is just moved from Restore.

dom.go
```diff
@@ -139,8 +139,37 @@ func (h *HTML) restoreText(prev *HTML) {
 	}
 }
 
-func (h *HTML) restoreHTML(prev *HTML) {
-	h.node = prev.node
+func (h *HTML) mutate(prev *HTML) {
+	// Check for compatible tag and mutate previous instance on match, otherwise start fresh
```
pdf: The following block is basically the block from old line 257 of Restore in this diff.

dom.go
```diff
@@ -175,6 +204,19 @@ func (h *HTML) restoreHTML(prev *HTML) {
 		}
 	}
 
+	// Dataset
```
pdf: This block was missing.

dom.go
```diff
@@ -110,12 +110,29 @@ type HTML struct {
...
 	eventListeners                  []*EventListener
 	children                        []ComponentOrHTML
+	new                             bool
 }
 
 // Node returns the underlying JavaScript Element or TextNode.
 func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
 
-func (h *HTML) restoreText(prev *HTML) {
+func (h *HTML) newNode() {
```
slimsag: nit: this doesn't return a new node as e.g. `new` (builtin) would, but rather it is initializing `h.node` so I might suggest `initNode` as being the name. Do you think that'd be a better name too, or disagree?

dom.go
```diff
@@ -124,8 +141,40 @@ func (h *HTML) restoreText(prev *HTML) {
 	}
 }
 
-func (h *HTML) restoreHTML(prev *HTML) {
-	h.node = prev.node
+func (h *HTML) mutate(prev *HTML) {
```
slimsag: I think that `reconcile` would be a better name for this, since it's reconciling state from a new `*HTML` against a prev / old `*HTML`. In fact, this was the original name before we renamed it to 'restore' in an attempt to shorten the (external) name. Since it's internal, though, I feel `reconcile` would do better than `mutate`. What do you think?

slimsag:
> This probably is the largest change I'd like to land, so I thought I'd try to get this out up front. I've split it up a little so that you can hopefully understand how I've arrived here (will squash before merge), however I'm hoping that the final changeset is not too large for comprehension.

This is a great PR (both in terms of code and how distinct each change is) -- let's _not_ squash merge (I prefer readers will be able to come to the same logical conclusion as the PR reviewer in general).

> The 10000ft view is that the majority of *HTML.Restore is redundant code, since *HTML.restoreHTML also performs these mutations.

If I understand correctly, you're using `h.new` to track whether or not the `*HTML` element here has just been created as opposed to being reconciled against an old one? i.e. to remove the 'redundant' code path?

> Additionally, the name restore is a misnomer here, because the code is actually about performing DOM mutations, and the code has never implemented Restorer, even though the comment suggests otherwise. I also can't see a reason for this code to be public.

I agree, the name is a misnomer and the comment is invalid / it should be private.

> Knowing whether we are mutating a compatible element from the previous render, or need to replace an incompatible element will allow us to make some more intelligent decisions about DOM manipulation in subsequent changes.

Are these 'intelligent decisions' made in this PR, or in the future? Can you elaborate on what you mean by this?

> I will note that we currently appear to be missing tests for all the HTML mutations (e.g. properties, attributes, styles, etc), so I'd like to get them added either before this is merged, or as part of this PR. One other thing to note is that this change highlighted that we were not previously mutating dataset between renders, only setting it on initial render.

This change is actually very easy for me to visually verify, I feel confident it won't break anything. So I won't block this PR on having these tests (unless you want to). Of course we should still get these in soon, and it will block any PRs I can't easily verify visually.

> Please let me know if this is comprehensible, though I'm not sure how I would make this one much smaller.

I just wanted to say this PR is actually very, very good. :) I appreciate you splitting this up _a lot_!

dom.go
slimsag: Actually, given every DOM function here is `create[Element|ElementNS|TextNode]` how about `createNode` ?

dom.go
```diff
@@ -142,9 +191,11 @@ func (h *HTML) restoreHTML(prev *HTML) {
 			h.node.Set(name, value)
 		}
 	}
-	for name := range prev.properties {
-		if _, ok := h.properties[name]; !ok {
-			h.node.Set(name, nil)
+	if !h.new {
+		for name := range prev.properties {
```
slimsag: I wonder if `h.new` is needed at all, actually. Won't `prev` always be `nil` in the case of new node creation? That would make all of our `range` statements here no-op, and then we can save ourselves the trouble of keeping the `h.new` state variable (and we save the memory + defer performance cost).

For readability sake, we can of course add a comment above these all (or above each one, whichever seems more clear) stating this.

dom.go
pdf: Sound fine to me.

dom.go
pdf: Fine by me.

dom.go
pdf: `prev` may be a populated, but incompatible tag.  However `h.new` will be valuable for subsequent PRs (comments to follow in the issue body).

pdf:
> If I understand correctly, you're using h.new to track whether or not the *HTML element here has just been created as opposed to being reconciled against an old one? i.e. to remove the 'redundant' code path?

Partially, more on this below.

> Are these 'intelligent decisions' made in this PR, or in the future? Can you elaborate on what you mean by this?

In subsequent PRs, knowing that we're mutating a compatible element will allow us to avoid calls to replace children, since the mutation satisfies the state change by itself, and it will have an impact on the keyed list implementation.  In the former case, calling replaceChild with the same node _should_ be efficient (noop) any way, but I'd prefer to avoid the call rather leave it up to the browser renderer.

> I feel confident it won't break anything. So I won't block this PR on having these tests (unless you want to).

I'll just address these separately then, agree we need them sooner than later.

> I just wanted to say this PR is actually very, very good. :) I appreciate you splitting this up a lot!

Excellent, I feel confident we can move forward in manageable chunks from this point.

dom.go
slimsag: Ah, you are right. Okay, let's go with the `h.new` approach you have here then.

slimsag:
> In subsequent PRs, knowing that we're mutating a compatible element will allow us to avoid calls to replace children, since the mutation satisfies the state change by itself, and it will have an impact on the keyed list implementation. In the former case, calling replaceChild with the same node should be efficient (noop) any way, but I'd prefer to avoid the call rather leave it up to the browser renderer.

I didn't think of that case, but very good point. I also strongly agree about avoiding the call rather than leaving it up to the browser.

slimsag:
Glad to merge this once those two naming comments are addressed (`mutate -> reconcile` and `newNode -> createNode`)

pdf:
Should be good to go :+1:

slimsag:
Can you reword that one commit `Rename *HTML.(r|R)estore* to *HTML.mutate, drop redundant code path` to `reconcile` as well? (easy to do with `git rebase -i`)

slimsag:
merged via CLI after renaming commit; also [updated the changelog](https://github.com/gopherjs/vecty/blob/master/docs/CHANGELOG.md#pre-v100-breaking-changes)

Thanks @pdf !

pdf:
Thanks @slimsag

PR #118, opened 2017-06-23
Add tests for element mutation

pdf:
Still missing are child mutations, and `value`/`checked` properties.  I'd also like to add tests for EventListener equality at some stage, though I'm not sure how to deal with that since they're called by JS, so I don't think we can mock the function argument.

### DIFF
--- a/dom_test.go
+++ b/dom_test.go
@@ -134,8 +134,390 @@ func TestHTML_reconcile_std(t *testing.T) {
 			t.Fatalf("got %q want %q", setNodeValue, want)
 		}
 	})
+	t.Run("properties", func(t *testing.T) {
+		cases := []struct {
+			name         string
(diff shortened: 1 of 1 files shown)


pdf:
This work did lead me to #119, which I have a change ready for if you agree that deletions are correct.

dom_test.go
```diff
@@ -134,8 +134,388 @@ func TestHTML_reconcile_std(t *testing.T) {
 			t.Fatalf("got %q want %q", setNodeValue, want)
 		}
 	})
-
-	// TODO(slimsag): test the restoreHTML code path
```
slimsag: Add:

```
// TODO: test (*HTML).reconcile child mutations, and value/checked properties
```

?

pdf: :+1:

PR #122, opened 2017-06-28
Fix rendering of child nodes for incompatible elements

### LINKED ISSUE #121: incorrect rendering when elements are added or removed
It seems that when re-rendering a component that has elements added or removed, something gets lost. Here is a minimal test case:

```
package main

import (
	"github.com/gopherjs/vecty"
)

type C struct {
	vecty.Core

	Show bool
}

func (c *C) Render() *vecty.HTML {
	return elem.Body(
		elem.Heading1(vecty.Text("Test Case")),
		vecty.If(c.Show, elem.Heading2(vecty.Text("foo"))),
		elem.Button(vecty.Text("Toggle"),
			event.Click(func(e *vecty.Event) {
				c.Show = !c.Show
				vecty.Rerender(c)
			})),
	)
}

func main() {
	c := &C{}
	vecty.RenderBody(c)
}
```

I'd expect the `<h2>Foo</h2>` to appear and disappear when clicking the "Toggle" button. 

What actually happens is that the DOM ends up looking like this, after clicking toggle (notice the missing text inside the H2:

```
<body><h1>Test Case</h1><h2></h2><button>Toggle</button></body>
```

I would expect that it would be:

```
<body><h1>Test Case</h1><h2>foo</h2><button>Toggle</button></body>
```

pdf:
Fixes #121 

I had this change in my previous code, but missed it when writing #117.  This change is required because new (incompatible) nodes always need their children appended.

I will try to get tests sorted out for child mutations over the weekend.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -259,7 +259,8 @@ func (h *HTML) reconcile(prev *HTML) {
 
 	// TODO better list element reuse
 	for i, nextChild := range h.children {
-		if i >= len(prev.children) {
+		// TODO(pdf): Add tests for h.new usage
+		if i >= len(prev.children) || h.new {
 			nextChildRender, skip := render(nextChild, nil)
 			if skip {
 				continue


dom.go
```diff
@@ -259,7 +259,7 @@ func (h *HTML) reconcile(prev *HTML) {
 
 	// TODO better list element reuse
 	for i, nextChild := range h.children {
-		if i >= len(prev.children) {
+		if i >= len(prev.children) || h.new {
```
slimsag: Could you add a simple `// TODO: write tests for h.new usage here` comment, just so I don't forget?

PR #124, opened 2017-06-30
Rendering a component must reconcile against previous DOM state

### LINKED ISSUE #123: (more) incorrect rendering with added/removed elements
Thanks for the quick response in #121, folks! I'm afraid we don't quite have the issue licked. Here is a test case that also fails:

```
package main

import (
	"log"

	"github.com/gopherjs/vecty"
)

type C struct {
	vecty.Core
	SubComponent *SubComponent
	Show         bool
}

func (c *C) Render() *vecty.HTML {
	if c.SubComponent == nil {
		c.SubComponent = &SubComponent{}
	}
	return elem.Body(
		elem.Heading1(vecty.Text("Test Case")),
		vecty.If(c.Show, c.SubComponent),
		elem.Button(vecty.Text("Toggle"),
			event.Click(func(e *vecty.Event) {
				c.Show = !c.Show
				vecty.Rerender(c)
			})),
	)
}

type SubComponent struct {
	vecty.Core
}

func (v *SubComponent) Render() *vecty.HTML {
	return elem.Div(
		elem.Heading2(vecty.Text("Subcomponent")),
		vecty.Text("Subcomponent Text"),
	)
}

func (v *SubComponent) Unmount() {
	log.Printf("never called -- should be?")
}

func main() {
	c := &C{Show: true}
	vecty.RenderBody(c)
}
```

In this case the text `"Subcomponent Text"` appears initially but when removing and re-adding the element it is not present. Also, it appears that the Unmount() function is not called. 

Documenting this for now, I'll do some more digging.

pdf:
The `renderComponent` function incorrectly presumed that `prevRender` from the component context must be the previous DOM state, however elements may have been shuffled/removed between component renders for persistent component pointers.

Fixes #123

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -335,7 +335,7 @@ func Rerender(c Component) {
 	if prevRender == nil {
 		panic("vecty: Rerender invoked on Component that has never been rendered")
 	}
-	nextRender, skip := renderComponent(c)
+	nextRender, skip := renderComponent(c, prevRender)
 	if skip {
 		return
 	}
@@ -379,7 +379,7 @@ func render(c ComponentOrHTML, prevRender *HTML) (h *HTML, skip bool) {
 		v.reconcile(prevRender)
 		return v, false
 	case Component:
-		return renderComponent(v)
+		return renderComponent(v, prevRender)
 	default:
 		panic(fmt.Sprintf("vecty: encountered invalid ComponentOrHTML %T", c))
 	}
@@ -388,7 +388,7 @@ func render(c ComponentOrHTML, prevRender *HTML) (h *HTML, skip bool) {
 // renderComponent handles rendering the given Component into *HTML. If skip ==
 // true is returned, the Component's SkipRender method has signaled the
 // component does not need to be rendered and h == nil is returned.
-func renderComponent(comp Component) (h *HTML, skip bool) {
+func renderComponent(comp Component, prevRender *HTML) (h *HTML, skip bool) {
 	// Now that we know we are rendering the component, restore friendly
 	// relations between the component and the previous component.
 	if comp != comp.Context().prevComponent {
@@ -419,7 +419,7 @@ func renderComponent(comp Component) (h *HTML, skip bool) {
 	}
 
 	// Restore the actual rendered HTML.
-	nextRender.reconcile(comp.Context().prevRender)
+	nextRender.reconcile(prevRender)
 
 	// Update the context to consider this render.
 	comp.Context().prevRender = nextRender
@@ -431,7 +431,7 @@ func renderComponent(comp Component) (h *HTML, skip bool) {
 // RenderBody renders the given component as the document body. The given
 // Component's Render method must return a "body" element.
 func RenderBody(body Component) {
-	nextRender, skip := renderComponent(body)
+	nextRender, skip := renderComponent(body, nil)
 	if skip {
 		panic("vecty: RenderBody Component.SkipRender returned true")
 	}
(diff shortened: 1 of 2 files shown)


domutil.go
```diff
@@ -1,12 +1,24 @@
...
+func parentNode(node jsObject) jsObject {
+	if node == nil {
+		panic("vecty: Cannot find parent of nil node")
+	}
+	parent := node.Get("parentNode")
+	if parent == nil {
+		panic("vecty: Missing parent node")
+	}
+
+	return parent
```
slimsag: remove blank line 11 here

domutil.go
```diff
@@ -1,12 +1,24 @@
 package vecty
 
+func parentNode(node jsObject) jsObject {
+	if node == nil {
+		panic("vecty: Cannot find parent of nil node")
+	}
+	parent := node.Get("parentNode")
+	if parent == nil {
+		panic("vecty: Missing parent node")
+	}
```
slimsag: Why do we need this function at all (instead of just invoking `node.Get("parentNode)` directly and assuming the result is non-nil. It'll panic just the same, and I hope that real user applications won't ever hit either of these cases.

slimsag:
I don't fully understand this change, so I'd like to ask a few questions.

1. Why is it that `Rerender` cannot utilize `comp.Context().prevRender`? 
2. If it's true that `Rerender` can't utilize `comp.Context().prevRender`, then what about other users of that variable? Can those also be incorrect, or no (why)?

I see the example code in https://github.com/gopherjs/vecty/issues/123#issue-239186034 but as far as I can tell from this PR, `Rerender` passes `c` and `c.Context().prevRender` as arguments to `renderComponent`, which then seems to directly use that `c.Context().prevRender` argument for reconciliation. So I don't see why `renderComponent` needs this new argument or how it fixes the issue.

On tests: I'm fine to merge this without tests once I understand it, but let's add a TODO in the test file that links to this PR for adding tests later.

domutil.go
pdf: Mostly because a panic on null is not very informative.

pdf:
In the case of `Rerender` with a top-level component, that works fine.  What we're looking at here though is a component (pointer) that's rendered as a child of the `Rerender` target (somewhere down the render tree).  If the component is a pointer that persists between renders, `Context().prevRender` for that component is not nil, but bears no relationship to what's actually in the DOM at that point (in the example, the child component's nodes have been removed from the DOM entirely, and the new render should cause it to replace some other elements, instead `reconcile` tries to manipulate nodes that don't exist).

The child component must be reconciled against the sibling DOM that the _parent_ rendered in the previous pass.

domutil.go
slimsag: Okay. I'd prefer to drop this part of the change if you're alright with that. I think that a package should really only panic (esp. with a `vecty:` prefix) if the user of the library did something obviously wrong. In this case.. I don't think this will be triggered due to a user error, so I'd prefer to keep the (less informative) nil pointer panic here.

slimsag:
Ok, that makes a lot more sense. Would I be correct in assuming that this also helps https://github.com/gopherjs/vecty/issues/78 then (although notably still retaining [this panic](https://github.com/pdf/vecty/blob/ad9067ef65c6476e306052797ce5fdc897b22c94/dom.go#L279-L281))?

> On tests: I'd definitely appreciate a hand working out how to sensibly test nested elements/components (in #29)

I'll see if I can find some time soon to look into this.

slimsag:
Once https://github.com/gopherjs/vecty/pull/124#discussion_r125157866 is addressed I will merge.

pdf:
I think #78 should be close to solved with this merge, combined with the rework of render/renderComponent/reconcile, but I've not tested sufficiently to call it conclusively.

slimsag:
Fair enough. I'll post my thoughts on that issue.

PR #129, opened 2017-07-24
Rename docs, examples to doc, example.

### LINKED ISSUE #128: examples directory should be renamed to example.
This is a really minor style nit, but I can't help but report it because this is one of the few remaining high quality repositories where I see a plural being used for a collection, which feels counter to idiomatic Go.

Reference: https://dmitri.shuralyov.com/idiomatic-go#use-singular-form-for-collection-repo-folder-name

This is already fixed in https://github.com/go-gl/example and https://github.com/neelance/graphql-go/tree/master/example.

shurcooL:
These directories are collections. They should use singular form. That is more consistent with the style the Go project uses.

Fixes #128.

@slimsag, do you know if I need to change [this line](https://github.com/gopherjs/vecty/blob/7d15bbeaa9172b71e239a83098d29ffc3b70590b/examples/todomvc/node_modules/todomvc-common/base.js#L133)? I can't tell if it's used anywhere, or what it's used for. It doesn't seem to be used in my testing. I tried both examples, and they worked successfully.


slimsag:
Thanks for sending this!

> @slimsag, do you know if I need to change this line? I can't tell if it's used anywhere, or what it's used for. It doesn't seem to be used in my testing. I tried both examples, and they worked successfully.

I'm not 100% certain, but I'm pretty sure that code comes from the todomvc project itself -- not anything vecty specific. If it breaks something, I'll fix in the future. :)

Feel free to send another PR to add yourself to the `CONTRIBUTORS` file!

shurcooL:
> but I'm pretty sure that code comes from the todomvc project itself -- not anything vecty specific.

That's what I suspect too. Thanks.

> Feel free to send another PR to add yourself to the CONTRIBUTORS file!

Thanks, but being listed in [git history](https://github.com/gopherjs/vecty/commits/master?author=shurcooL) is good enough for me for now!

PR #130, opened 2017-08-07
Lifecycle design changes (persistent component instances)

### LINKED ISSUE #115: Design: Determine how to manage component lifecycle
In the course of #114, issues with current methods for instantiating child components were identified that will result in hard to reason about behaviour for users.

This issue is to track design discussions on methods to alleviate these issues.

My preference is to remove the need for the `Restorer` interface entirely by somehow retaining a reference to a child component between parent renders, though I'm not sure exactly what this looks like at this stage.

slimsag:
Fixes #115

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -245,12 +245,11 @@ func (h *HTML) reconcile(prev *HTML) {
 		h.node.Set("innerHTML", h.innerHTML)
 	}
 
-updatingChildren:
 	// TODO better list element reuse
 	for i, nextChild := range h.children {
 		// TODO(pdf): Add tests for h.new usage
 		if i >= len(prev.children) || h.new {
-			nextChildRender, skip := render(nextChild, nil, nil)
+			nextChildRender, skip := render(nextChild, nil)
 			if skip {
 				continue
 			}
@@ -259,28 +258,24 @@ updatingChildren:
 		}
 
 		prevChild := prev.children[i]
-		prevChildRender, ok := prevChild.(*HTML)
-		if !ok {
-			prevChildRender = prevChild.(Component).Context().prevRender
-
-			// Keep prevChild Component as our child, so that the next time we
-			// are here prevChild will be the same. We do this because
-			// prevChild is the persistent component instance.
-			h.children[i] = prevChild.(Component)
+		prevChildRender := assertHTML(prevChild)
+		nextChildRender, skip := render(nextChild, prevChild)
+		if nextChildRender == prevChildRender {
+			panic("vecty: next child render must not equal previous child render (did the child Render illegally return a stored render variable?)")
 		}
-
-		nextChildRender, skip := render(nextChild, prevChild, prevChildRender)
+		if prevComponent, ok := prevChild.(Component); ok {
+			if nextComponent, ok := nextChild.(Component); ok && matchComponent(prevComponent, nextComponent) {
+				h.children[i] = prevChild
+			}
+		}
 		if skip {
-			continue updatingChildren
+			continue
 		}
 		replaceNode(nextChildRender.node, prevChildRender.node)
 	}
 	for i := len(h.children); i < len(prev.children); i++ {
 		prevChild := prev.children[i]
-		prevChildRender, ok := prevChild.(*HTML)
-		if !ok {
-			prevChildRender = prevChild.(Component).Context().prevRender
-		}
+		prevChildRender := assertHTML(prevChild)
 		removeNode(prevChildRender.node)
 		if u, ok := prevChild.(Unmounter); ok {
 			u.Unmount()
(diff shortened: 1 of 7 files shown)


slimsag:
Also see this branch which has one additional commit, which I don't intend to merge (for obvious reasons). It's a test / example file I used for developing this: https://github.com/gopherjs/vecty/tree/lifecycle-test

dom.go
```diff
@@ -418,20 +454,26 @@ func renderComponent(comp Component, prevRender *HTML) (h *HTML, skip bool) {
...
-	// Restore the actual rendered HTML.
+	// Reconcile the actual rendered HTML.
 	nextRender.reconcile(prevRender)
 
 	// Update the context to consider this render.
-	comp.Context().prevRender = nextRender
-	comp.Context().prevComponent = comp
-	comp.Context().prevRenderComponent = doCopy(comp)
+	//
+	// TODO: In the event that this really is the previous component instance,
```
slimsag: I'm like 80% certain this is the cause of the issue mentioned in this TODO comment of my test file: https://github.com/gopherjs/vecty/commit/7fc6935b2ca7722641a4e2fa90dc30d06623697f#diff-59654adfef9097fef9e9fede9315019aR45

The overarching problem here is that, now that we have persistent component references, it's actually not always obvious where the proper `Context()` really is. Someone will need to sit down and rethink how to fix this issue here + hopefully make the `Context` storage location a bit easier to think about..

slimsag:
@pdf here is what I've managed to get. It's EOD Sunday for me, and I'm not sure when I'll have time again to dive into Vecty code (probably a few weeks).

In an effort to not block you from making changes, I suggest:

1. **If this looks mostly good to you**, I can merge it and you can work from there.
2. **If you think it needs more work before merging**, I can close this PR and you can create a new one based on this one.
3. **If you think this is all wrong for some reason**, or that there's a better approach to take, I can close and you can create a brand new PR. :)

Just let me know which you prefer ^^

pdf:
Will be next weekend before I can review and make a determination, thanks @slimsag

dom.go
pdf: I suspect the behaviour that the comment refers to is a bug I'm already aware of, where a rerender gets queued for a child component that in the mean time gets removed or replaced in the DOM by a parent rerender.  If this is indeed the same bug, it will be fixed by lifecycle event changes (mount/unmount) that I already have in the pipeline.

slimsag: Actually I don't think that is it that. The bug here is that:

1. Imagine a two-component setup: a parent, and its child.
1. `Rerender(thePrevParent)` is called, where `thePrevParent` is the 'previous' component instance (one the user has a pointer to).
1. `renderComponent(thePrevParent, ...)` will be called and so the 'previous' component instance (`thePrevParent.Context()`) will be updated on lines 467-469 below.
1. When its children are walked, it will effectively invoke `renderComponent(theNextChild, thePrevChildContext, thePrevChildRender)`, with `theNextChild` being the newly created component instance -- NOT the persistent / previous component instance. So the 'next' component instance (`theNextChild.Context()`) will be updated on lines 467-469.

To be entirely honest, this is hard for me to reason about, and that's part of the problem^. The troublesome bit here is that... it's not clear whether or not the 'next' or 'previous' `Component.Context` has the information and which needs to be updated.

I'll play around with this some more to see if I can resolve this issue / remove the ambiguity..

slimsag: Ok, fixing the above actually turned out to be quite easy. I just needed to make the persistent component instance always be the source of truth for Context information.

I've done this in https://github.com/gopherjs/vecty/pull/130/commits/df133f3e38d8d450bd1567b4c77e65f8d9d1f09f and it fixes that TODO (see e391b1bb41223afa8b5cf322f58f269bc65e64d2)

slimsag:
Sorry for the last-minute change BTW, hope that isn't too confusing. ^^  let me know what you think..

pdf:
No problem, however I think there are some bad assumptions made, particularly in the latest commit - I don't believe it accounts for the possibility that a child may change types between renders, and there are no checks for component type equality.

I've taken your commits that clean up the `Restore`-related code, and your `copyProps` func, plus some of your improved comments into an implementation that I believe is a little simpler, and more accurate in #131.  I would certainly like to add some tests though.

Let me know what you think, and we can continue either here or there, based on your thoughts.

pdf:
I wasn't quite ready for this to be merged to master, as there are some outstanding things to discuss, mostly around what I'll cover in #132.

slimsag:
@pdf apologies for the premature merge, I figured that me and you both felt pretty confident about this implementation and, even if not complete due to things like #132, merging would be OK given Vecty hasn't had a 1.0 release yet. I should have waited for your go ahead, though, I'll do that next time! If desireable, we can temporarily revert out of `master`.

pdf:
No problem, I should have been explicit about my opinion of the code state.  My concern is mostly for people trying to take Vecty for a spin and running into the inconsistent and hard to diagnose behaviour that this change may exhibit currently.

shurcooL:
The changelog says:

> Properties should be denoted via ``` `vecty:"prop"` ``` struct field tags.

What are properties and what are non-properties? How do I know if a field should be marked with that struct field tag?

pdf:
Fields that are updated from the parent need to be tagged:

```
type InnerComponent struct {
  vecty.Core
  ParentSet bool `vecty:"prop"`
  LocalState bool
}

func (c *InnerComponent) Render() *vecty.HTML {
  return elem.Div(
    vecty.If(c.ParentSet, elem.Span(`true updated from parent`)),
  )
}

type OuterComponent struct {
  vecty.Core
  flip bool
}

func (c *OuterComponent) Render() *vecty.HTML {
  c.flip = !c.flip
  return elem.Div(
    &InnerComponet{ParentSet: c.flip},
  )
}
```

PR #131, opened 2017-08-13
Lifecycle design changes (2) (persistent component instances)

### LINKED ISSUE #115: Design: Determine how to manage component lifecycle

pdf:
Second pass at implementing persistent component references.  Main differences to #130 are checks for compatible child types, simplified render args, and added child stability when using `vecty.If()`.

The requirement for the latter does illustrate possibly surprising behaviour for people rendering lists with changing children, that I was expecting to encounter for this implementation because persistence relies on component position amongst siblings to be stable. However, adding the keyed child implementation from #112 and friends in addition to these changes should make it easy to avoid this pitfall for more complex use-cases.  I do wonder thought whether we shouldn't consider trying to use some sort of internal `Noop` HTML type that gets skipped for rendering output, but is retained as a HTML child on the struct, so that we don't litter the DOM with `<noscript>` elems.

I still need to look at how to write tests for this stuff.

Fixes #115

### DIFF
--- a/markup.go
+++ b/markup.go
@@ -170,7 +170,19 @@ func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
 	if cond {
 		return List(markup)
 	}
-	return nil
+	var result []MarkupOrComponentOrHTML
+	for _, m := range markup {
+		if m == nil {
+			continue
+		}
+		switch m.(type) {
+		case *HTML, Component:
+			result = append(result, Tag("noscript"))
+		default:
+			return nil
+		}
+	}
+	return List(result)
 }
 
 // UnsafeHTML is Markup which unsafely sets the inner HTML of an HTML element.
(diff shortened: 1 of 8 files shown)


markup.go
```diff
@@ -170,7 +170,19 @@ func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
...
+			continue
+		}
+		switch m.(type) {
+		case *HTML, Component:
+			result = append(result, Tag("noscript"))
+		default:
+			return nil
+		}
+	}
+	return List(result)
```
slimsag: I'm going to submit this file's changes as a separate PR so that we can discuss a bit further.

PR #132, opened 2017-08-14
Return <noscript> from vecty.If() for consistent child length/order

slimsag:


slimsag:
@pdf my first thought about this change is: I wonder if Vecty's new persistent component instances differ substantially enough from React to warrant this change.

For example, in React you would write if logic using just JS's `if` and I cannot imagine that React would be able to / would try to appropriately map the missing child to a `<noscript>` or no-op as you mentioned in the PR description of #131 

I am hesitant to make `If` do more magic behind the scenes than it advertises up front, so I would like to figure out 1. if this is really needed, and 2. what the equivalent code might look like in React.

markup.go
```diff
@@ -170,7 +170,19 @@ func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
 	if cond {
 		return List(markup)
 	}
-	return nil
+	var result []MarkupOrComponentOrHTML
+	for _, m := range markup {
+		if m == nil {
+			continue
```
slimsag: I think this `continue` here would cause a bug where this would produce one less element to end up in `result`, no?

```
If(false, Tag("p"), nil, Tag("p")) // produces -> [Tag("noscript"), Tag("noscript")]
```

Instead of the three noscript elements it would produce when called with `true`:

```
If(true, Tag("p"), nil, Tag("p")) // produces -> [Tag("noscript"), nil, Tag("noscript")]
```

markup.go
```diff
@@ -170,7 +170,19 @@ func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
...
+		}
+		switch m.(type) {
+		case *HTML, Component:
+			result = append(result, Tag("noscript"))
+		default:
+			return nil
```
slimsag: I'm not sure why the default here is `return nil`, because I think this would lead to invalid behavior.

The argument type to `If` is `...MarkupOrComponentOrHTML`, and you only handle `*HTML, Component` explicitly above. That means this `default` case would happen in the case of invalid `MarkupOrComponentOrHTML` (in which case we should panic), or in the case of `Markup`.

If this is for the case of `Markup` type, then I think that:

```
If(false, Tag("p"), If(true, Tag("p"))) // produces -> [Tag("noscript")]
If(true, Tag("p"), If(true, Tag("p"))) // produces -> [Tag("noscript"), If(true, Tag("p"))]
```

Which surely can't be what we want here?

markup.go
```diff
@@ -170,7 +170,19 @@ func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
...
 	}
-	return nil
+	var result []MarkupOrComponentOrHTML
+	for _, m := range markup {
+		if m == nil {
+			continue
+		}
+		switch m.(type) {
+		case *HTML, Component:
+			result = append(result, Tag("noscript"))
```
slimsag: I think in this case, it is Vecty internally which handles `nil`, so we would just write:

```
result = append(result, nil)
```

i.e. we shouldn't use `noscript` here directly, but rather let Vecty handle the `nil` -> `noscript` translation itself.

slimsag:
@pdf take a look at my comments/questions here whenever you have some spare time :)

pdf:
Some general comments, then I'll address the inline comments.

> For example, in React you would write if logic using just JS's if and I cannot imagine that React would be able to / would try to appropriately map the missing child to a <noscript> or no-op as you mentioned in the PR description of #131

Honestly, I have no idea how React manages this - the renderer(s) are pretty complex chunks of code.  However, JSX in React doesn't work in remotely the same way as Vecty's rener code, for example it doesn't have conditional statements.  I suspect they can be a lot more clever about tracking elements in their virtual DOM, as they're parsing the JSX and strictly interpolating variable content.  Consider contrived JSX:

```jsx
render() {
  let variableContent = '<span>variable</span>';
  if (condition) {
    variableContent = undefined;
  }
  return (
    <div>
      <h1>heading</h1>
      {variableContent}
      <ReactComponent prop="blah" />
    </div>
  )
}
```

When rendering this, React may not know whether anything will be rendered for `variableContent`, or how many elements might be there, but it knows there is interpolation to be performed there that can be tracked, so whatever gets rendered there, `ReactComponent` always comes after that content, and the mounted `ReactComponent` can be maintained across renders.

The equivalent Vecty code:

```go
func (c *MyComponent) Render() *HTML {
  return elem.Div(
    elem.Header1(`header`),
    vecty.If(condition, elem.Span("variable")),
    &VectyComponent{Prop: "blah"}, // This component will get reset every time `condition` changes
  )
}
```

Now, because component persistence depends on child position, unless `vecty.If()` maintains a placeholder matching the number of elements it would normally render, `VectyComponent` will lose persistence every time the `condition` changes.

There are other problems that I haven't worked out how to solve yet, like the fact that a user can render variable element length markup from a function call or slice var.

markup.go
pdf: Vecty performs this translation in the case of a component `Render` return value being nil.  Possibly we could handle this within `List.Apply()` instead of `If` though, which would give better coverage.  I couldn't see a sane way of incorporating this into `apply()` in `markup.go` though.

markup.go
pdf: I don't think `true` is a valid type for `MarkupOrComponentOrHTML` anyway, is it?

The idea was that only `Component` and `HTML` should return tags, but `Markup` should just return `nil` immediately.  If users are mixing the latter with the former, they'll find out they screwed up when their condition matches I guess, sames as if they just put the wrong mix outside `If`.

markup.go
pdf: The curly part is when the first element in the slice is a `nil`, since we don't know whether we're dealing with `Component`/`HTML`, or with `Markup`.  I suppose the correct thing to do might be to iterate over the whole slice before doing anything - if we encounter any `Markup`, return `nil`, otherwise just return a slice of `Tag("noscript")` with the same length as the argument, though I'm not sure what the correct behaviour should be for a list of `nil`s.

pdf:
I'm thinking as an alternative to this particular change about having `List` be sort of lazy-evaluated inside `reconcile`, so that it maintains a single slot on `HTML.children` that gets expanded into the DOM when rendered.

markup.go
slimsag: > Vecty performs this translation in the case of a component Render return value being nil.

Good point. You are correct about this.

markup.go
slimsag: > I don't think true is a valid type for MarkupOrComponentOrHTML anyway, is it?

I'm not sure how you got `true` as being a valid `MarkupOrComponentOrHTML` type from my example code above. Maybe take a second look?

boolean values like `true` certainly are not a valid `MarkupOrComponentOrHTML`, my comment was 100% about the type `Markup` not being handled correctly.

> The idea was that only Component and HTML should return tags, but Markup should just return nil immediately. If users are mixing the latter with the former, they'll find out they screwed up when their condition matches I guess, sames as if they just put the wrong mix outside If.

What about this?

```
var list List
list = append(list, Tag("p"))
list = append(list, Tag("p"))
If(false, list)
```

Wouldn't the user expect the `If` in that case to return two `<noscript>` tags, but with this implementation it would instead just return `nil`?

pdf: Sorry, reading comprehension failure on my part.  I don't really want to return <noscript> tags here at all, this was just a stop-gap change that highlights the problem with the current persistence implementation wrt changing sibling counts.  I must admit I hadn't considered the `Tag()` markup though.

slimsag:
Thanks for the detailed explanation in https://github.com/gopherjs/vecty/pull/132#issuecomment-322133780 I appreciate that a lot.

> I'm thinking, as an alternative to this particular change, about having List be sort of lazy-evaluated inside reconcile, so that it maintains a single slot on HTML.children that gets expanded into the DOM when rendered.

I was thinking about something very similiar to this. The nice thing about this approach would be that `If` can remain a helper function rather than magical, i.e.:

```Go
// If returns nil if cond is false, otherwise it returns the given markup.
func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
	if cond {
		return List(markup)
	}
	return nil
}
```

AND it actually makes slightly more sense because currently the only way to apply `MarkupOrComponentOrHTML` to an `*HTML` pointer is via `List`, so it already has some quite specialized / `*HTML` specific logic in `apply`.

pdf:
In the negative case for `If`, it would return an empty list, rather than `nil` in my thinking - this way the child slot is occupied but doesn't render anything.

slimsag:
Isn't a `nil` child in `(*HTML).children` doing the same thing? (taking up a child slot and not rendering anything), or is that because of some code to remove nil children I am forgetting about?

slimsag:
(either way, an empty list sounds like a good starting point)

pdf:
`func apply()` in `markup.go` skips `nil`s.

slimsag:
right... it may be enough to just remove that. In fact, I think that would cause a similar issue to this one if a user provided e.g. a list of components sometimes containing nil, it probably makes more sense to not skip over them there.

pdf:
The problem (similarly evidenced by this issue) is determining when it's valid to modify `HTML.children`, because we absolutely do not want to do that when receiving `nil` in a context where that `nil` may be replaced not by an element producer, but with a property/attribute/listener/etc.  Happy to hear suggestions on how this might be handled sanely, but I'm not seeing it.

I believe there's value in the `List` approach either way - it gets us much closer to the React example above - the List can act similarly to the interpolated variable, isolating an arbitrary chunk of elements into a separate context, giving us stable children regardless of what gets rendered in the List.

I need to mull this over a little, hopefully I'll get to have a poke at it on the weekend.

slimsag:
> The problem (similarly evidenced by this issue) is determining when it's valid to modify HTML.children, because we absolutely do not want to do that when receiving nil in a context where that nil may be replaced not by an element producer, but with a property/attribute/listener/etc. Happy to hear suggestions on how this might be handled sanely, but I'm not seeing it.
>
> Whilst skipping nils may be convenient in some circumstances, it may actually make more sense to make them illegal to resolve the above problems, since If is available to handle optional children.

Perhaps you could write a small code snippet to demonstrate this problem? I've read over this statement multiple times, but I fear I am just not understanding it correctly. Here is what I am thinking in code terms:

Today, `MarkupOrComponentOrHTML` looks like this (note that it does not mention `nil` today, which means the code comment is actually inaccurate)

```Go
// MarkupOrComponentOrHTML represents one of:
//
//  Markup
//  Component
//  *HTML
//
type MarkupOrComponentOrHTML interface{}
```

It seems to me that the situations in which we want to avoid `(MarkupOrComponentOrHTML)(nil)` is when it is a `Markup` or `Component`, but not when it is an `*HTML`. So I propose we do not change the above `MarkupOrComponentOrHTML` implementation, but rather just start treating `MarkupOrComponentOrHTML((*HTML)(nil))` differently.

I think that `apply` could then be written like this:

```diff
func apply(m MarkupOrComponentOrHTML, h *HTML) {
-	if m == nil {
-		return
-	}
	switch m := m.(type) {
	case nil:
		return
	case Markup:
		m.Apply(h)
	case Component:
		h.children = append(h.children, m)
	case *HTML:
-		if m == nil {
-			return
-		}
		h.children = append(h.children, m)
	default:
		panic(fmt.Sprintf("vecty: invalid type %T does not match MarkupOrComponentOrHTML interface", m))
	}
}
```

This would mean that any `(Markup)(nil)` or `(Component)(nil)` would always be ignored by the list, and `(*HTML)(nil)` would be added to `h.children` as you would expect.

The reason I think this makes more sense is because today you can return `nil` from a component `Render` method, and it'll render `<noscript>`.. but today you cannot do `If(true, nil)` and have it render a `<noscript>`, instead the list order changes.


> I believe there's value in the List approach either way - it gets us significantly closer in confidence to the React example above - the List can act similarly to the interpolated variable, isolating an arbitrary chunk of elements into a separate context, giving us stable children regardless of what gets rendered in the List (and we'd want to apply the logic recursively, so that nested Lists get the same treatment).

Completely agree.

pdf:
I'll try to explain my thoughts on `nil` as they currently stand:

```go
type OtherComponent struct {
  vecty.Core
}

func (c *OtherComponent) Render() *vecty.HTML {
  return elem.Div()
}

type MyComponent struct {
  vecty.Core
  b bool
}

  c.b = !c.b
  return elem.Div(
    vecty.If(c.b, &OtherComponent{}),
    &OtherComponent{}, // <- loses persistence every render
  )
}
```

Since `If` returns `nil`, the `nil` case in your `apply()` is hit, resulting in every second render missing a child.  Even if we modify `If` to work around this by returning a `List` as I suggested earlier, any time we encounter a concrete `nil`, we're in trouble:

```go
  c.b = !c.b

  var flipper vecty.MarkupOrComponentOrHTML
  if c.b {
    flipper = &OtherComponent{}
  }
  return elem.Div(
    flipper,
    &OtherComponent{}, // <- loses persistence every render
  )
}
```

Same result, same for, say, `Tag("span", ...)` being switched out for `nil`.

The problem is that I don't think we can be much smarter in handling a concrete `nil` type, because we can't possibly know whether the `nil` might be swapped for something that should take up a child slot, or for some `Markup` like `Property`/`Attribute`/etc that shouldn't.

Taking the above example, where the `*HTML.children` count for the `Div` flips between 1 and 2, the below has a constant child count of 1, but there's no way to tell in `apply()` which case we're encountering:

```go
  c.b = !c.b

  var flipper vecty.MarkupOrComponentOrHTML
  if c.b {
    flipper = vecty.Attribute("k", "v")
  }
  return elem.Div(
    flipper,
    &OtherComponent{}, // <- maintains persistence across renders
  )
}
```

If we make `nil` illegal with a panic (making `MarkupOrComponentOrHTML`'s comment accurate, and requiring the use of `If` for conditionals) these inconsistencies go away, and combined with the `List` idea we should have stable children in all scenarios.

I think it's fine though to allow a `nil` return from `Component.Render()` and have it handled specially.

slimsag:
Mostly makes sense now, I appreciate the detailed explanation. One question this raises for me with your recommended change:

```Go
  c.b = !c.b

  flipper := vecty.If(c.b, vecty.Attribute("k", "v"))
  return elem.Div(
    flipper,
    &OtherComponent{}, // <- maintains persistence across renders
  )
}
```

How would `flipper` avoid becoming a `List{}` here, thus taking up one child slot? Or does `If` try to look at the type of the arguments to determine whether or not `List{}` should consume one slot?

---

One thing I dislike about this approach is that users _must_ now use `vecty.If` in place of their normal `if` `else` logic statements, whereas previously `vecty.If` was seen as more of a nice helper to use inside `elem.Foo(...)` blocks. For example:

```Go
	var x vecty.MarkupOrComponentOrHTML
	if c.b {
		x = vecty.Attribute("k", "v")
	} else if (c.d) {
		x = vecty.Attribute("k", "v2")
	}
	return elem.Div(
		x,
		&OtherComponent{},
	)
}
```

Now must become:

```Go
	x := vecty.If(c.b, vecty.Attribute("k", "v"))
	if !c.b && c.d {
		x = vecty.Attribute("k", "v2")
	}
	return elem.Div(
		x,
		&OtherComponent{},
	)
}
```

One can imagine much more convoluted situations, but maybe this is not possible to solve / just something we have to accept.

pdf:
> How would flipper avoid becoming a List{} here, thus taking up one child slot? Or does If try to look at the type of the arguments to determine whether or not List{} should consume one slot?

A `List` would always take up a slot on `*HTML.children`, but `vecty.Attribute()` is stored on `*HTML.attributes`, not `*HTML.children`.

---

> One thing I dislike about this approach is that users must now use `vecty.If`

I mean, technically they don't, all I'm really proposing there is that they'll get a panic if their value is ever `nil`, and `vecty.If` will guard against this.  If you wanted to use `vecty.If` for your example though, you could do it like so:

```go
	return elem.Div(
		vecty.If(c.b, vecty.Attribute("k", "v")),
		vecty.If(!c.b, vecty.Attribute("k", "v2")),
		&OtherComponent{},
	)
}
```

pdf:
My weekend starts tomorrow, so @slimsag if you have any input on my recent comments before I start making changes, they would be most welcome

slimsag:
Thanks for the ping!

I assume that with the snippet I presented earlier:

```
	var x vecty.MarkupOrComponentOrHTML
	if c.b {
		x = vecty.Attribute("k", "v")
	} else if (c.d) {
		x = vecty.Attribute("k", "v2")
	}
	return elem.Div(
		x,
		&OtherComponent{},
	)
}
```

Translating this snippet to become:

```diff
func (c *MyComponent) Render() *vecty.HTML {
-	var x vecty.MarkupOrComponentOrHTML
+	var x = vecty.List{}
	if c.b {
		x = vecty.Attribute("k", "v")
	} else if (c.d) {
		x = vecty.Attribute("k", "v2")
	}
	return elem.Div(
		x,
		&OtherComponent{},
	)
}
```

would be the most trivial way to translate this code and ensure the `nil` panic is never encountered, because `vecty.List` is `Markup` and `vecty.Attribute` is also `Markup` -- but is this code valid? I do not think it is, because in one case it is a list taking up a slot on `*HTML.children` and in another two cases it is stored in `*HTML.attributes`, not `*HTML.children`, just as you said:

> A List would always take up a slot on *HTML.children if I can make that work, but vecty.Attribute() is stored on *HTML.attributes, not *HTML.children.

So I assume that _somehow_ you intend to prevent such a situation from happening -- but I'm not really sure.

---

> I'm certainly open to other suggestions on how to allow nil to work if you have them.

FWIW At this point, I don't think `nil` can work.

---

> One option that may be worth looking at is trying to find some way to separate children from stuff like attributes/properties/etc.

I REALLY like this idea a lot! I wish I'd thought of this sooner.

```Go
  return elem.Div(
    vecty.Attribute("a", "v"),
    vecty.Property("p", "v"),
    vecty.Style(...),
    vecty.Children(
      elem.Div(),
      nil,
      &OtherComponent{},
    ),
  )
```

I think that perhaps aside from a potential snafu with regenerating the `elem` subpackage (because the source it is generated from is often changed / the generator may be broken), this could be a *very* good idea. If the generator is broken, please don't hesitate to just manually change `elem/` and I'll fix the generator later.

There is an added benefit here, which is that this kind of sets a precedent / standardized style that `Children` are in one location in code, typically the end, whereas attributes/styles/etc are at the front of the code.

I don't think it'd be very hard at all to implement the above, in fact here's a play snippet that does it -- I'm sure other little things will need changes -- but it should be quite easy with this approach:

https://play.golang.org/p/DI6bH5yUVZ

I'll do my best to be active/online throughout tomorrow + my weekend to respond to questions / review PRs / etc. 

**Tl;DR: I think your last approach there is VERY good, and can be easily done by changing `Tag()` to take just `Markup` and adding a `Children` function that returns `Markup`.**

pdf:
Thanks for the input - I do think `Children` (or the reverse) is likely give us a clean and easy to reason about result.  The only downside with `Children` (besides the large API breakage) is that it often results in significant nesting boilerplate:

```go
  return elem.Div(
    vecty.Children(
      vecty.Div(
        vecty.Children(
          vecty.Span(
            vecty.Children(
              ...,
            ),
          ),
        ),
      ),
    ),
  )
}
```

slimsag:
Yes, that's true. I am not concerned about API breakage, since we're not 1.0 yet. But the nesting boilerplate is ugly, I admit.

Did you have any thoughts on the first part of my message?

pdf:
> Yes, that's true. I am not concerned about API breakage, since we're not 1.0 yet. But the nesting boilerplate is ugly, I admit.

This is the main reason I included the opposite approach where non-children are wrapped, since you don't get nesting for them, though I'm not sure what terminology to use for the wrapper func.

> Did you have any thoughts on the first part of my message?

Yes, mixing child-producing values and non-child-producing values for a `MarkupOrComponentOrHTML` variable is a problem I've not yet worked out how to solve, though it _would_ be solved by the `Children` (or opposite) approach.

slimsag:
> This is the main reason I included the opposite approach where non-children are wrapped, since you don't get nesting for them, though I'm not sure what terminology to use for the wrapper func.

Understood. I didn't make sense of that the first time around when reading this. That being the case, my current opinion is that is the best approach of the three choices we have.

Terminology: how about renaming `Markup` today to `Applyer`, and using `Markup` for the wrapper func?

e.x. https://play.golang.org/p/GvJQm80OAw

Your snippet becomes:

```
  return elem.Div(
    vecty.Markup(
      vecty.Attribute("a", "v"),
      vecty.Property("p", "v"),
      vecty.Style(...),
    ),
    elem.Div(),
    nil,
    &OtherComponent{},
  )
}
```

pdf:
Sounds alright to me, I'll have a go at implementing these changes over the weekend.  Not sure of the best way to describe the usage of `Markup()` in documentation though.

slimsag:
I can follow up any code PR with a documentation PR of that sounds helpful

pdf:
Sure, I'll put something together, and we can go from there. :+1:

pdf:
Didn't quite get there - the `nil`/`List` implementation turned out to be somewhat challenging, hopefully I'll get to sort it out next weekend.

slimsag:
Closing in favor of #134

PR #134, opened 2017-08-25
Stable Children

pdf:
This PR aims to provide stable element children, to enable persistent `Component`s.

This is not final code, as I'm not really happy with some of this, however I believe it is at least correctly functional unless I've missed something, and I'd like to start getting some feedback.  I think we absolutely need to do something about testing here though - this got rather complicated.

Supercedes #132

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: go
 go:
-  - 1.8.x
+  - 1.9.x
 install:
   - nvm install node
   - npm install -g source-map-support
--- a/dom.go
+++ b/dom.go
@@ -291,7 +291,7 @@ func (h *HTML) reconcile(prev *HTML) {
 // Tag returns an HTML element with the given tag name. Generally, this
 // function is not used directly but rather the elem subpackage (which is type
 // safe) is used instead.
-func Tag(tag string, m ...MarkupOrComponentOrHTML) *HTML {
+func Tag(tag string, m ...MarkupOrChild) *HTML {
 	h := &HTML{
 		tag: tag,
 	}
@@ -304,7 +304,7 @@ func Tag(tag string, m ...MarkupOrComponentOrHTML) *HTML {
 // Text returns a TextNode with the given literal text. Because the returned
 // HTML represents a TextNode, the text does not have to be escaped (arbitrary
 // user input fed into this function will always be safely rendered).
-func Text(text string, m ...MarkupOrComponentOrHTML) *HTML {
+func Text(text string, m ...MarkupOrChild) *HTML {
 	h := &HTML{
 		text: text,
 	}
--- a/dom.go
+++ b/dom.go
@@ -64,6 +64,8 @@ type Unmounter interface {
 //
 //  Component
 //  *HTML
+//  List
+//  nil
 //
 // If the underlying value is not one of these types, the code handling the
 // value is expected to panic.
@@ -245,12 +247,26 @@ func (h *HTML) reconcile(prev *HTML) {
 		h.node.Set("innerHTML", h.innerHTML)
 	}
 
+	h.reconcileChildren(prev, nil)
+}
+
+// reconcileChildren reconciles children of the current HTML against a previous
+// render's DOM nodes.
+func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) {
 	// TODO better list element reuse
 	for i, nextChild := range h.children {
 		// TODO(pdf): Add tests for h.new usage
 		if i >= len(prev.children) || h.new {
+			if nextChildList, ok := nextChild.(List); ok {
+				nextChildList.reconcile(h.node, insertBefore, nil)
+				continue
+			}
 			nextChildRender, skip := render(nextChild, nil)
-			if skip {
+			if skip || nextChildRender == nil {
+				continue
+			}
+			if insertBefore != nil {
+				h.node.Call("insertBefore", nextChildRender.node, insertBefore.node)
 				continue
 			}
 			h.node.Call("appendChild", nextChildRender.node)
@@ -258,9 +274,42 @@ func (h *HTML) reconcile(prev *HTML) {
 		}
 
 		prevChild := prev.children[i]
-		prevChildRender := extractHTML(prevChild)
+		var prevChildRender *HTML
+		// If the previous child was not a list, extract the previous child
+		// render.
+		if _, isList := prevChild.(List); !isList {
+			prevChildRender = extractHTML(prevChild)
+		}
+
+		// If the previous child render was nil, try to find the next DOM node
+		// in the previous render so that we can insert this child at the
+		// correct location.
+		if prevChildRender == nil {
+			for j := i + 1; j < len(prev.children) && insertBefore == nil; j++ {
+				if prevChildList, ok := prev.children[j].(List); ok {
+					insertBefore = prevChildList.firstHTML()
+				} else {
+					insertBefore = extractHTML(prev.children[j])
+				}
+			}
+		}
+
+		// If the next child is a list, reconcile its elements in-place, and
+		// we're done.
+		if nextChildList, ok := nextChild.(List); ok {
+			nextChildList.reconcile(h.node, insertBefore, prevChild)
+			continue
+		}
+
+		// If the previous child was a list, remove the list elements from the
+		// previous render, since we no longer have a list.
+		if prevChildList, ok := prevChild.(List); ok {
+			prevChildList.remove(h.node)
+		}
+
+		// Determine the next child render.
 		nextChildRender, skip := render(nextChild, prevChild)
-		if nextChildRender == prevChildRender {
+		if nextChildRender != nil && prevChildRender != nil && nextChildRender == prevChildRender {
 			panic("vecty: next child render must not equal previous child render (did the child Render illegally return a stored render variable?)")
 		}
 
@@ -276,16 +325,94 @@ func (h *HTML) reconcile(prev *HTML) {
 		if skip {
 			continue
 		}
-		replaceNode(nextChildRender.node, prevChildRender.node)
+
+		// Perform the final reconciliation action for nextChildRender and
+		// prevChildRender. Replace, remove, insert or append the DOM nodes.
+		switch {
+		case nextChildRender == nil && prevChildRender == nil:
+			continue // nothing to do.
+		case nextChildRender != nil && prevChildRender != nil:
+			replaceNode(nextChildRender.node, prevChildRender.node)
+		case nextChildRender == nil && prevChildRender != nil:
+			removeNode(prevChildRender.node)
+		case nextChildRender != nil && prevChildRender == nil:
+			if insertBefore != nil {
+				h.node.Call("insertBefore", nextChildRender.node, insertBefore.node)
+				continue
+			}
+			h.node.Call("appendChild", nextChildRender.node)
+		default:
+			panic("vecty: internal error (unexpected switch state)")
+		}
 	}
+	h.removeChildren(prev)
+}
+
+// removeChildren removes child elements from the previous render pass that no
+// longer exist on the current HTML children.
+func (h *HTML) removeChildren(prev *HTML) {
+	if prev == nil {
+		return
+	}
+	// Every previous child that h.children does not have in common.
 	for i := len(h.children); i < len(prev.children); i++ {
 		prevChild := prev.children[i]
+		if prevChildList, ok := prevChild.(List); ok {
+			// Previous child was a list, so remove all DOM nodes in it.
+			prevChildList.remove(h.node)
+			continue
+		}
 		prevChildRender := extractHTML(prevChild)
+		if prevChildRender == nil {
+			continue
+		}
 		removeNode(prevChildRender.node)
 		if u, ok := prevChild.(Unmounter); ok {
 			u.Unmount()
 		}
 	}
+}
+
+// List represents a list of components or HTML.
+type List []ComponentOrHTML
+
+// reconcile reconciles the List against the DOM node in isolation. If
+// insertBefore is non-nil, the List elements will be inserted into the DOM
+// before that node.
+func (l List) reconcile(node jsObject, insertBefore *HTML, prev ComponentOrHTML) {
+	nextHTML := &HTML{node: node, children: l}
+	switch c := prev.(type) {
+	case List:
+		nextHTML.reconcileChildren(&HTML{node: node, children: c}, insertBefore)
+	default:
+		if prev == nil {
+			nextHTML.reconcileChildren(&HTML{node: node}, insertBefore)
+			return
+		}
+		nextHTML.reconcileChildren(&HTML{node: node, children: []ComponentOrHTML{prev}}, insertBefore)
+	}
+}
+
+// firstHTML returns the content of the first element from which a *HTML can be
+// extracted. Returns nil if not found.
+func (l List) firstHTML() (h *HTML) {
+	for _, v := range l {
+		if listChild, ok := v.(List); ok {
+			h = listChild.firstHTML()
+		} else {
+			h = extractHTML(v)
+		}
+		if h != nil {
+			return h
+		}
+	}
+	return nil
+}
+
+// remove the List's elements from the DOM node
+func (l List) remove(node jsObject) {
+	nextHTML := &HTML{node: node}
+	nextHTML.removeChildren(&HTML{node: node, children: l})
 }
 
 // Tag returns an HTML element with the given tag name. Generally, this
@@ -422,6 +549,8 @@ func render(next, prev ComponentOrHTML) (h *HTML, skip bool) {
 	case Component:
 		// Cases 4, 5, and 6 above.
 		return renderComponent(v, prev)
+	case nil:
+		return nil, false
 	default:
 		panic(fmt.Sprintf("vecty: encountered invalid ComponentOrHTML %T", next))
 	}
--- a/markup.go
+++ b/markup.go
@@ -34,7 +34,7 @@ func (l *EventListener) StopPropagation() *EventListener {
 	return l
 }
 
-// Apply implements the Markup interface.
+// Apply implements the Applyer interface.
 func (l *EventListener) Apply(h *HTML) {
 	h.eventListeners = append(h.eventListeners, l)
 }
@@ -45,38 +45,40 @@ type Event struct {
 	Target *js.Object
 }
 
-// MarkupOrComponentOrHTML represents one of:
+// MarkupOrChild represents one of:
 //
-//  Markup
 //  Component
 //  *HTML
+//  List
+//  nil
+//  MarkupList
 //
 // If the underlying value is not one of these types, the code handling the
 // value is expected to panic.
-type MarkupOrComponentOrHTML interface{}
+type MarkupOrChild interface{}
 
-func apply(m MarkupOrComponentOrHTML, h *HTML) {
+func apply(m MarkupOrChild, h *HTML) {
 	if m == nil {
 		return
 	}
 	switch m := m.(type) {
-	case Markup:
+	case MarkupList:
 		m.Apply(h)
-	case Component:
-		h.children = append(h.children, m)
-	case *HTML:
+	case List:
+		m.apply(h)
+	case Component, *HTML:
 		if m == nil {
 			return
 		}
 		h.children = append(h.children, m)
 	default:
-		panic(fmt.Sprintf("vecty: invalid type %T does not match MarkupOrComponentOrHTML interface", m))
+		panic(fmt.Sprintf("vecty: invalid type %T does not match MarkupOrChild interface", m))
 	}
 }
 
-// Markup represents some type of markup (a style, property, data, etc) which
+// Applyer represents some type of markup (a style, property, data, etc) which
 // can be applied to a given HTML element or text node.
-type Markup interface {
+type Applyer interface {
 	// Apply applies the markup to the given HTML element or text node.
 	Apply(h *HTML)
 }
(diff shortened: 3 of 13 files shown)


markup.go
```diff
@@ -154,26 +150,49 @@ func (m ClassMap) Apply(h *HTML) {
...
-func If(cond bool, markup ...MarkupOrComponentOrHTML) MarkupOrComponentOrHTML {
+func If(cond bool, markup ...ComponentOrHTML) ComponentOrHTML {
+	if cond {
+		return ChildList(markup)
+	}
+	return nil
+}
+
+// MarkupIf returns nil if cond is false, otherwise it returns the given markup.
+func MarkupIf(cond bool, markup ...Applyer) Applyer {
```
pdf: I don't really like having this, but no better options came to mind.

markup.go
```diff
@@ -154,26 +150,49 @@ func (m ClassMap) Apply(h *HTML) {
...
 
+// Markup wraps a list of Applyer which is individually
+// applied to an HTML element or text node.
+func Markup(m ...Applyer) MarkupList {
+	return MarkupList(m)
+}
+
+// List wraps a list of Markup, Component, or HTML which is individually
+// applied to an HTML element or text node.
+func List(m ...ComponentOrHTML) ChildList {
```
pdf: It's possible for a user to shoot themselves in the foot here due to the empty interface types, and get a panic at render time, like so:

```go
var blah []ComponentOrHTML
vecty.List(blah)
```
When they should instead do:
```go
var blah []ComponentOrHTML
vecty.List(blah...)
```

pdf: We could alternatively just retain:
```go
type List []ComponentOrHTML
```
instead of `ChildList`, but it felt odd that `Markup` was a function, whereas `List` was a type, so they would require different syntax to use.

pdf:
There are also certainly render efficiency improvements that can be made once we get correctness nailed.

markup.go
slimsag: > We could alternatively just retain: ...

The reason `Markup` is a function is because the most common use case for `Markup` is something like this:

```
return elem.Div(
	vecty.Markup(
		vecty.Style("float", "right"),
	),
	...
}
```

So that users don't have to write:

```
return elem.Div(
	[]vecty.Markup{
		vecty.Style("float", "right"),
	},
	...
}
```

Which wouldn't really match the style of other things inside a component's `Render` method.

---

In contrast to the above, `List` is _not_ commonly used inline inside a `elem.Foobar()` function call. For example, I do not really expect users to write:

```
return elem.Div(
	vecty.List(
		vecty.Text("Hello"),
		vecty.Text("World"),
	),
}
```

Because they could instead just write that as:

```
return elem.Div(
	vecty.Text("Hello"),
	vecty.Text("World"),
}
```

What I would expect users to do is use `vecty.List` in place of a regular `[]ComponentOrHTML`, for example in code like https://github.com/pdf/vecty/blob/e71f4c2cda8a9fcd493e78d1034bb9befed0ab2e/example/todomvc/components/pageview.go#L179-L185, because writing `ComponentOrHTML` is tedious. IMHO for this use case, `vecty.List()` being a function seems more ambiguous as to what it really does.

---

Because of the above, I would say that we should just retain:

```
type List []ComponentOrHTML
```

Also, with the above, they can't shoot themselves in the foot with at least that specific example:

```
var blah []ComponentOrHTML
vecty.List(blah)
```

because that is now a type conversion. Of course, there are other various ways they could still shoot themselves in the foot, though.

markup.go
```diff
@@ -154,26 +150,49 @@ func (m ClassMap) Apply(h *HTML) {
...
+		if a == nil {
+			continue
+		}
+		a.Apply(h)
 	}
 }
 
+// Markup wraps a list of Applyer which is individually
+// applied to an HTML element or text node.
+func Markup(m ...Applyer) MarkupList {
```
slimsag: I think the return type here should just be `Applyer`.

I think the description should be simplified to just:

```
// Markup returns an applyer that applies all of the given markup.
```

markup.go
slimsag: I've thought about this for a bit, and haven't come up with any great options either. I'll keep thinking about it.

markup.go
```diff
@@ -154,26 +150,49 @@ func (m ClassMap) Apply(h *HTML) {
...
 // applied to an HTML element or text node.
-type List []MarkupOrComponentOrHTML
+type MarkupList []Applyer
 
-// Apply implements the Markup interface.
-func (l List) Apply(h *HTML) {
-	for _, m := range l {
-		apply(m, h)
+// Apply implements the Applyer interface.
+func (m MarkupList) Apply(h *HTML) {
```
slimsag: I personally wonder if having this exported makes sense or not.

I think the number of situations where someone will use it is quite low, and they could use `var markup []vecty.Applyer` & `vecty.Markup(markup...)` for such cases.

My instinct says this should be unexported until we have good evidence that it would be fairly useful to people, this way we keep our exported API footprint smaller.

markup.go
```diff
@@ -47,36 +47,32 @@ type Event struct {
...
-	if m == nil {
-		return
-	}
 	switch m := m.(type) {
-	case Markup:
-		m.Apply(h)
-	case Component:
-		h.children = append(h.children, m)
-	case *HTML:
+	case MarkupList:
```
slimsag: This case would just become `Applyer`

markup.go
```diff
@@ -47,36 +47,32 @@ type Event struct {
 
 // MarkupOrComponentOrHTML represents one of:
 //
-//  Markup
+//  MarkupList
 //  Component
 //  *HTML
+//  ChildList
```
slimsag: `nil` should also be added here.

markup.go
```diff
@@ -47,36 +47,32 @@ type Event struct {
...
 //
 // If the underlying value is not one of these types, the code handling the
 // value is expected to panic.
 type MarkupOrComponentOrHTML interface{}
```
slimsag: The name now does not reflect the actual type. But of course `MarkupListOrComponentOrHTMLOrChildListOrNil` is not a great name.

Perhaps shorter more abstract name would do us good, something like `MarkupOrChild`.

markup.go
```diff
@@ -47,36 +47,32 @@ type Event struct {
 
 // MarkupOrComponentOrHTML represents one of:
 //
-//  Markup
+//  MarkupList
```
slimsag: Couldn't this be `Applyer` instead?

dom.go
```diff
@@ -347,6 +483,15 @@ func extractHTML(e ComponentOrHTML) *HTML {
 	}
 }
 
+func isChildList(e ComponentOrHTML) bool {
+	switch e.(type) {
+	case ChildList:
+		return true
+	default:
+		return false
+	}
```
slimsag: Please use simpler syntax:

```
_, ok := e.(ChildList)
return ok
```

dom.go
```diff
@@ -245,22 +247,69 @@ func (h *HTML) reconcile(prev *HTML) {
...
 
+	h.reconcileChildren(prev, nil)
+}
+
+// reconcileChildren reconciles children of the current HTML against a previous
+// render's DOM nodes.
+func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) {
+	if prev == nil {
+		prev = &HTML{}
+	}
```
slimsag: What does this do?

dom.go
```diff
@@ -245,22 +247,69 @@ func (h *HTML) reconcile(prev *HTML) {
...
+			if skip || nextChildRender == nil {
+				continue
+			}
+			if insertBefore != nil {
+				h.node.Call("insertBefore", nextChildRender.node, insertBefore.node)
+				continue
+			} else {
+				h.node.Call("appendChild", nextChildRender.node)
 				continue
 			}
```
slimsag: Prefer less indention, 273 already has `continue` so just remove `else`

dom.go
```diff
@@ -276,18 +325,105 @@ func (h *HTML) reconcile(prev *HTML) {
...
+		idx, _ := l[exceptIndex].(ChildList).firstHTML()
+		l[exceptIndex].(ChildList).removeExcept(node, idx)
+		children = append(l[:exceptIndex], l[exceptIndex+1:]...)
+	default:
+		children = append(l[:exceptIndex], l[exceptIndex+1:]...)
+	}
+	nextHTML := &HTML{node: node}
+	nextHTML.removeChildren(&HTML{node: node, children: children})
+}
+
```
slimsag: All of the code in this file above this line is quite complex, and I don't understand most of it currently (I've only had ~15 mins to look at it, my lunch break is up now).

I'll have more time to look at this later today or tomorrow, but I think we should be able to simplify a lot of this or make it more clear what is going on. Hopefully my other comments will at least let us solidify the public API.

markup.go
pdf: The same applies foot-shooting caveat applies to the Markup function though:

```go
var markup []Applyer
vecty.Markup(markup) // wrong
```

The awkwardness though was that if users have a children var of `[]ComponentOrHTML`, and a markup var of `[]Applyer`, the syntax to use them is different if we retain the old `List` type:

```go
var children []ComponentOrHTML
var markup []Applyer

vecty.Markup(markup...)
vecty.List(children)
```

I... think I'm okay with changing `List` though...

markup.go
pdf: I specifically _don't_ want `Applyer` to be a valid type at the top level, that would mess up our ability to work with `nil` correctly.

markup.go
pdf: Yep, this was left exposed by accident.

markup.go
pdf: Yeah, I noticed this after I left my other comments, but didn't want to force-push just for that.

dom.go
pdf: Allows calling `reconcileChildren` on the first pass where the prev render is `nil`, sames as early in `*HTML.reconcile()`.

dom.go
pdf: ack

dom.go
pdf: Thanks @slimsag, open to suggestions - as I said in the OP, I do not expect this to be final code, but a first functional pass.  The list handling is necessarily somewhat complex though.

dom.go
pdf: ack

markup.go
pdf: There is still an issue that I forgot to mention - swapping `nil` for a `MarkupList` between renders will cause us problems - I actually considered not exporting `MarkupList` to avoid this, but having unexported return types can be annoying.

markup.go
pdf: See https://github.com/gopherjs/vecty/pull/134#discussion_r135370011

markup.go
pdf: See https://github.com/gopherjs/vecty/pull/134#discussion_r135370011

markup.go
slimsag: Let's ignore the foot-shooting portion: I agree it occurs in both cases.

---

I'm aware of the awkwardness you describe, but you're assuming that users will be converting their own `children` list via `vecty.List(children)`, why? i.e. with this change, you are saying that code previously written as:

```
var children vecty.List
children = append(children, foobar)
return Div(
    children,
)
```

should instead become:

```
var children []ComponentOrHTML
children = append(children, foobar)
return Div(
    vecty.List(children),
)
```

Sure, the above could just be `var children vecty.ChildList` -- but that is kind-of my point: I don't see in which scenario people would use this `vecty.List` helper function?

pdf: ack, I'll change this in the next pass.

dom.go
slimsag: Understandable :) I'm going to take a more deep look at this code tonight or tomorrow and offer more concrete suggestions than just "the code is complex" :P I think this PR is a really good starting point!

markup.go
slimsag: Having an unexported type be returned sounds a bit hacky.

> I specifically don't want Applyer to be a valid type at the top level, that would mess up our ability to work with nil correctly.

I'd really like to understand this more, could you point me to relevant code where this being an `Applyer` would cause trouble? Or provide a small example demo'ing the problem?

pdf: `nil` now occupies a child slot, so anything that returned `nil` in place of `Applyer` would break component persistence.

If `Applyer` is valid at the top level (in which case, there's no point having a `Markup` function, and `Applyer` could just stay named `Markup`):
```go
func (c *MyComponent) OptionalMarkup() vecty.Applyer {
        if c.condition {
                return vecty.Property("key", "value")
        }

        return nil
}

        return elem.Div(
                c.OptionalMarkup(), // breaks persistence
                &OtherComponent{},
        )
}
```

We currently have a similar problem with `MarkupList`:
```go
func (c *MyComponent) OptionalMarkup() vecty.MarkupList {
        if c.condition {
                return vecty.Markup(vecty.Property("key", "value"))
        }

        return nil
}

        return elem.Div(
                c.OptionalMarkup(), // breaks persistence
                &OtherComponent{},
        )
}
```

However, that becomes impossible if `MarkupList` is unexported, since it cannot be used as a return value, so we will never be able to receive a concrete `nil` in it's place, and it forces the use of `vecty.Markup()` for every case.

slimsag: Ahhh I see, understood. I had forgotten that `MarkupList` must effectively sit side-by-side with `Component` and `*HTML`, and that nil `*HTML` occupies a child slot.

This would not be the case of we forced users to supply Markup always:

```
func Div(markup vecty.Markup, children ...vecty.ComponentOrHTML) *vecty.HTML {
    ...
}
```

But understandably that would be a worse API for situations like `elem.Paragraph(vecty.Text("hello world!"))`.

I understand why you proposed an unexported `MarkupList` above, now, and agree that sounds best.

markup.go
pdf: Sounds fair to me.

pdf:
I have the current review notes queued (with some other minor style changes), but I'll wait for comments on the list handling before doing a rebased push.

slimsag:
@pdf feel free to force push now. It'd be good for me to get a view of it with all pending comments fixed.

dom.go
pdf: This may actually be unnecessary now, as some of the calling code has changed, will review.

pdf:
`golint` doesn't like the unexported return value.

pdf:
I think we have two options:
1. keep the unexported return and suppress the lint warning as I've done here, by moving to `gometalinter` (not a terrible idea anyway, as we get some extra checks there)
2. use two arguments to separate markup from children, though this makes the API awkward

markup.go
```diff
@@ -154,26 +151,43 @@ func (m ClassMap) Apply(h *HTML) {
...
+func If(cond bool, markup ...ComponentOrHTML) MarkupOrChild {
 	if cond {
 		return List(markup)
 	}
 	return nil
 }
 
-// UnsafeHTML is Markup which unsafely sets the inner HTML of an HTML element.
+// MarkupIf returns nil if cond is false, otherwise it returns the given markup.
+func MarkupIf(cond bool, markup ...Applyer) Applyer {
```
slimsag: Still an open question to me if there is a better API than this. But I won't block this PR on that, it's something we can improve in the future if we find a better alternative.

.travis.yml
```diff
@@ -6,20 +6,19 @@ install:
   - nvm install node
   - npm install -g source-map-support
   - go get -u github.com/gopherjs/gopherjs
-  - go get -u github.com/golang/lint/golint
+  - go get -u gopkg.in/alecthomas/gometalinter.v1
```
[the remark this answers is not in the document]
slimsag: I've just realized a second benefit to this approach. Imagine a situation where a user has some very complex markup for many different logical cases (e.g. themes or something like that) and they want to split it out into separate functions:

```Go
func GetTheme(name string) []Applyer {
	...
}
```

Callers would have to use it like this:

```Go
return elem.Div(
	vecty.Markup(foo.GetTheme("dark")...),
	...
)
```

Which, if you're a complete beginner, does look a bit strange. This is because `GetTheme` can't return the actual value of `vecty.Markup` to the user of the package. In contrast, with it being exported as suggested above, they could write:

```Go
func GetTheme(name string) vecty.MarkupList {
	return vecty.Markup(getTheme("dark")...)
}
```

And then callers may just use it like:

```Go
return elem.Div(
	foo.GetTheme("dark"),
	...
)
```

slimsag:
Sorry for the delay on reviewing this @pdf -- took me longer to get around to this than I thought it would.

I know today (Sun for me) is a weekday for you, so in addition to the (many) inline comments I've made above, I have also addressed each and every one of them and sent it as a PR to your branch here via https://github.com/pdf/vecty/pull/1

If I assume that you mostly agree with the comments I've made above, then I see three remaining issues for this PR before merging:

1. This comment https://github.com/gopherjs/vecty/pull/134#discussion_r135441646
2. Testing
3. Updating `docs/CHANGELOG.md`

On the topic of testing, I have been thinking about it a fair bit. I have some good ideas on how to handle this, by essentially creating some nice helpers in our test suite and then describing results effectively just as a list of HTML / Component operations that occurred. It will probably be a few more weeks before I can get to this, however, and I feel that this PR is a clear improvement over the current status quo (pending resolution of the comment #1 mentioned above) so I'd be fine merging this PR without tests once you feel confident about this being the right approach as well.

slimsag:
Will need to update the CI to use Go 1.9 as well since it is now failing due to GopherJS requiring latest Go version.

.travis.yml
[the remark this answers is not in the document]
pdf: Just to clarify a few things about `gometalinter`.  It's been around for a long time, and is pretty widely used.

1. It _vendors_ checks so that versions are stable, and installs always succeed, but they've had plenty of changes upstreamed to those projects.  megacheck is able to get away with not vendoring since all the checks are written by the same author.
2. Honestly, I'm not sure whether new linters will be added to v1 or not, but being explicit about what checks are run seems fine to me either way, otherwise new checks may unexpectedly break your CI.
3. The config file is optional, I just used it to make the CI config clearer, since we special-case `golint` for the generated code.  Choosing which checks to run is already something every project does, whether the checks are bundled under a single runner or not, by choosing what to enable in the CI run - the box is already open ;-)
4. Not exactly, it just makes it easy to run multiple existing checks via a single executable call - this is particularly useful for editor integration.

megacheck is very new by comparison (though I'm sure the quality is high), wrapping some of Dominik's other tools, though they've been available via gometalinter for a long time.  I'm not really phased either way, but I thought some of the criticisms were a little hard on gometalinter.

---

As for the API suggestion here, I like it :+1:

dom.go
```diff
@@ -276,22 +321,109 @@ func (h *HTML) reconcile(prev *HTML) {
...
+	switch {
+	case exceptIndex < 0:
+		children = l
+	case isList(l[exceptIndex]):
+		// TODO(pdf): I'm not sure I like the assumption here that we want to
+		// retain the first element of sub-lists, but not sure if there's a
+		// better method for handling this.
+		idx, _ := l[exceptIndex].(List).firstHTML()
+		l[exceptIndex].(List).removeExcept(node, idx)
+		children = append(l[:exceptIndex], l[exceptIndex+1:]...)
```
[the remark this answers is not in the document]
pdf: When we render `List`s, we do so recursively - all elements of all nested lists end up rendered as children of the first outer list's parent node, so when operating in the reverse we need to ensure that we recurs all child lists there too.

I wrote this logic before I was certain whether we would support a `nil` child though, so before final merge, I think I may just remove `exceptIndex` altogether and recursively remove all list elements, then we'll append/insert the new child.  It's an extra DOM operation, but I think that's okay for clarity.  Will need to do some testing though.

pdf:
@slimsag no problem, thanks for the review - the changes look good to me.

I'm very keen to get this tested, but improvements to the test harness would make the job significantly easier.  Though this is a significant breaking change, I think once it's finalised this should go in regardless, as the current code on master is prone to unexpected behaviour, and I think this is a sensible approach to resolving that.

dom.go
```diff
@@ -276,22 +321,109 @@ func (h *HTML) reconcile(prev *HTML) {
...
+	h.removeChildren(prev)
+}
+
+// removeChildren removes child elements from the previous render pass that no
+// longer exist on the current HTML children.
+func (h *HTML) removeChildren(prev *HTML) {
+	if prev == nil {
+		return
 	}
 	for i := len(h.children); i < len(prev.children); i++ {
```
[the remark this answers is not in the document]
pdf: Because in the case where there are no prev.children, this will panic, I've omitted this change.

pdf:
I merged your changes with a couple of minor modifications, added the change mentioned in https://github.com/gopherjs/vecty/pull/134#discussion_r135463934, and cleaned up the history.  I think this is good for final review now.

I tried bumping the Go version on Travis, but looks like they haven't finished rolling out 1.9 support yet.

slimsag:
Try tip version for Travis? May work.

shurcooL:
`go1.9.x` syntax seems to be working in Travis now. See https://travis-ci.org/go-interpreter/wagon/jobs/269323333.

pdf:
Thanks for the heads-up @shurcooL, looks like we still need the `GOROOT` hack though.

pdf:
And travis has fixed the GOROOT issue now too, no more 1.9 hacks required.

dom.go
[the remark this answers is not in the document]
slimsag: Got it, I missed that. Thanks for omitting!

slimsag:
> Though this is a significant breaking change, I think once it's finalised this should go in regardless, as the current code on master is prone to unexpected behaviour, and I think this is a sensible approach to resolving that.

I agree highly.

I'll do one final look through of this change + write up the doc/CHANGELOG.md as soon as I can, possibly tonight (Thur) or Saturday morning.

slimsag:
Also, do you have any final concerns about this PR? Feel it's good to go in its current state?

pdf:
It occurred to me the other night that we could now get rid of the `<noscript>` output for `nil` returns from component renders.  I'd like to include that too, and it shouldn't be too much effort, though I might not finish it this morning before work.

pdf:
Actually, allowing `nil` renders from components will have some implications for `vecty.Rerender()` that I may not have time to resolve this weekend, so with a couple of changes to support that approach at a later, I'm going to call this PR now, and we can address `<noscript>` in a subsequent PR.

slimsag:
Sounds good.

slimsag:
Alright, I've just now had time to do a final pass over this PR. Nothing sticks out to me, so I am going to go ahead and merge this and update the `doc/CHANGELOG.md` directly following.

slimsag:
Changelog updated, see https://github.com/gopherjs/vecty/blob/master/doc/CHANGELOG.md#sept-2-2017-pr-134-major-breaking-change

pdf:
Changelog looks good.  Is it worth making mention on slack or somewhere of the breakage?

slimsag:
Yep :) I mentioned in Slack channel as well

PR #138, opened 2017-09-04
handle nil *HTML in render

slimsag:
This change fixes a regression introduced by PR #134.

The basic issue is that when `nil` is encountered, it is treated correctly as a nil render,
but when `(*HTML)(nil)` is encountered, it is not. I discovered this while porting
Go-Package-Store (which uses Vecty) over to the new API.

The following code effectively demonstrates the issue:

```Go
package main

import (
	"github.com/gopherjs/vecty"
)

func main() {
	vecty.RenderBody(&PageView{})
}

type PageView struct {
	vecty.Core
}

func (p *PageView) Render() *vecty.HTML {
	return elem.Body(
		// this works:
		//nil,

		// this does not:
		p.returnsNil(),
	)
}

func (p *PageView) returnsNil() *vecty.HTML {
	return nil
}
```

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -543,6 +543,9 @@ func copyProps(src, dst Component) {
 func render(next, prev ComponentOrHTML) (h *HTML, skip bool) {
 	switch v := next.(type) {
 	case *HTML:
+		if v == nil {
+			return nil, false
+		}
 		// Cases 1, 2 and 3 above. Reconcile against the prevRender.
 		v.reconcile(extractHTML(prev))
 		return v, false
--- a/dom.go
+++ b/dom.go
@@ -543,9 +543,6 @@ func copyProps(src, dst Component) {
 func render(next, prev ComponentOrHTML) (h *HTML, skip bool) {
 	switch v := next.(type) {
 	case *HTML:
-		if v == nil {
-			return nil, false
-		}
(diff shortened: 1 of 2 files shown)


slimsag:
@pdf

pdf:
What about adding a `nil` check inside the case at https://github.com/gopherjs/vecty/blob/8eca1d5d9a4612f21920781987b0c58155969ac8/markup.go#L64?

slimsag:
Yes, that works as well. Does the latest commit look good to you? If so I'll squash merge

pdf:
I'd add it to the compound case, then it handles a `nil` `Component` or `List` return value too.

slimsag:
- `nil` obviously would not change, it'd still be `nil`.
- `(List)(nil)` should already be handled elsewhere where we interact with `List`, via `range` etc.
- `Component` is an `interface` and as such can't run into this problem. For example https://play.golang.org/p/SSg2vbLLk8 

So the only question is about readability of the code, I think?

slimsag:
(I confirmed changing the `returnsNil` return type in PR description example code to `Component` and `List` also works just fine with the code as-is)

pdf:
You are indeed correct, LGTM then.

Issue #79, opened 2017-01-27
Expose a way to do work on requestAnimationFrame


shazow:
Seems like a fairly natural thing that a lot of apps do (especially games), and a better practice than using `time.Tick`.

shazow:
Handy reference from @influx6 on the slack: https://github.com/influx6/faux/blob/master/raf/raf.go

slimsag:
Once #152 is merged, `vecty.Rerender` on a component uses `requestAnimationFrame` internally. That is: to animate a component, use `vecty.Rerender`.

Other potential uses of `requestAnimationFrame` are not covered (i.e. we don't literally expose that function), but I think that is fine / not in the scope for Vecty.

PR #139, opened 2017-09-05
Fix Unmounter behaviour and implement Mounter

### LINKED ISSUE #75: Support for fine-grained internal event handling (element creation, re-creation, removal)

### LINKED ISSUE #85: Incorrect Unmounter behavior in certain situations

### LINKED ISSUE #135: CustomEvent for when the actual DOM node is created
Hi,

I'm creating this issue because I had a need to get the actual DOM node for one of my components. The use case was:

"When rendering a list of nodes, I want to scroll to the last element on that list".

This is can be done using "domNode.scrollIntoView()".

So, in order to enabled that I changed a few things on "vecty"

on "package event":
```
// DOMCreated is called when the given dom node is created but it might not
// be inserted into the DOM.
//
// Use RAF from withing this handler to ensure your node is present on the document.
func DOMCreated(listener func(ev *vecty.Event)) *vecty.EventListener {
	return &vecty.EventListener{
		Name:     "domcreated",
		Listener: listener,
	}
}

// RAF executes step on the next frame
func RAF(step func()) {
	js.Global.Call("requestAnimationFrame", step)
}
```
on "package vecty" (dom.go)
```
// Incompatible node, start fresh
if prev == nil {
	prev = &HTML{}
}
h.createNode()
defer func() {
	h.new = false

	// fire an event on the dom object
	doc := js.Global.Get("document")
	createEv := doc.Call("createEvent", "HTMLEvents")
	createEv.Call("initEvent", "domcreated", false, true)
	h.node.Call("dispatchEvent", createEv)
}()
```

Then in my component function:
```
func (m *MyComponent) scrollToView(ev *vecty.Event) {
	event.RAF(func() {
		ev.Target.Call("scrollIntoView")
	})
}

func (m *MyComponent) Render() *vecty.HTML {
	return elem.Article(
		event.DOMCreated(m.scrollToView),
		... // other nodes/attributes/etc...
	)
}
```

Using the RAF solution isn't pretty, but given how vecty works I think it'll work all the times and this is also a standard so we can trust that it will be present on the browsers.

I didn't included any polyfill or browser detection but this is just an idea.

There is no branch/fork since I want to know if this is a desired feature and if the implementation is OK.

Thanks in advance.

pdf:
Had a few hours spare tonight, and this has been on my mind, so I took a first pass at it.  I think this is in pretty decent shape if you get a chance to look over it @slimsag.

Fixes #75, fixes #85, fixes #135

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -12,6 +12,7 @@ import (
 type Core struct {
 	prevRenderComponent Component
 	prevRender          *HTML
+	unmounted           bool
 }
 
 // Context implements the Component interface.
@@ -50,6 +51,14 @@ type Component interface {
 type Copier interface {
 	// Copy returns a copy of the component.
 	Copy() Component
+}
+
+// Mounter is an optional interface that a Component can implement in order
+// to receive component mount events.
+type Mounter interface {
+	// Mount is called after the component has been mounted, after the DOM node
+	// has been attached.
+	Mount()
 }
 
 // Unmounter is an optional interface that a Component can implement in order
@@ -129,13 +138,13 @@ func (h *HTML) reconcileText(prev *HTML) {
 	}
 }
 
-func (h *HTML) reconcile(prev *HTML) {
+func (h *HTML) reconcile(prev *HTML) []Mounter {
 	// Check for compatible tag and mutate previous instance on match, otherwise start fresh
 	switch {
 	case prev != nil && h.tag == "" && prev.tag == "":
 		// Compatible text node
 		h.reconcileText(prev)
-		return
+		return nil
 	case prev != nil && h.tag != "" && prev.tag != "" && h.tag == prev.tag && h.namespace == prev.namespace:
 		// Compatible element node
 		h.node = prev.node
@@ -242,12 +251,12 @@ func (h *HTML) reconcile(prev *HTML) {
 		h.node.Set("innerHTML", h.innerHTML)
 	}
 
-	h.reconcileChildren(prev, nil)
+	return h.reconcileChildren(prev, nil)
 }
 
 // reconcileChildren reconciles children of the current HTML against a previous
 // render's DOM nodes.
-func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) {
+func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts []Mounter) {
 	// TODO better list element reuse
 	for i, nextChild := range h.children {
 		// TODO(pdf): Add tests for node equality
@@ -256,9 +265,13 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) {
 				nextChildList.reconcile(h.node, insertBefore, nil)
 				continue
 			}
-			nextChildRender, skip := render(nextChild, nil)
+			nextChildRender, skip, mounters := render(nextChild, nil)
 			if skip || nextChildRender == nil {
 				continue
+			}
+			pendingMounts = append(pendingMounts, mounters...)
+			if m, ok := nextChild.(Mounter); ok {
+				pendingMounts = append(pendingMounts, m)
 			}
 			if insertBefore != nil {
 				h.node.Call("insertBefore", nextChildRender.node, insertBefore.node)
@@ -303,7 +316,7 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) {
 		}
 
 		// Determine the next child render.
-		nextChildRender, skip := render(nextChild, prevChild)
+		nextChildRender, skip, mounters := render(nextChild, prevChild)
 		if nextChildRender != nil && prevChildRender != nil && nextChildRender == prevChildRender {
 			panic("vecty: next child render must not equal previous child render (did the child Render illegally return a stored render variable?)")
 		}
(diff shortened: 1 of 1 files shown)


pdf:
At some stage I would like to move the `dom_utils.go` functions to methods on `*HTML`, and integrate these lifecycle events into the DOM manipulation calls, to ensure they're always called as the codebase evolves.  I elected not to do so in this PR though, so as to make the changeset easier to parse.

dom.go
```diff
@@ -52,11 +52,19 @@ type Copier interface {
 	Copy() Component
 }
 
+// Mounter is an optional interface that a Component can implement in order
+// to receive component mount events.
+type Mounter interface {
+	// Mount is called after the component has been mounted, after the DOM node
+	// has been attached.
```
slimsag: replace `, after the DOM node` with `, i.e. after the DOM node` since we're explaining what it means for a component to be mounted. There isn't an important distinction between a component being mounted and a DOM node being attached.

slimsag:
Sorry I haven't had a chance to look at this fully yet. I should have some time this weekend.

pdf:
No problem.  Was thinking I might start on keyed children next, unless you have thoughts on other priorities?

slimsag:
@pdf sounds good. Any chance you could create issues for the remaining items that were in your original PR (https://github.com/gopherjs/vecty/pull/112/files) that are not yet in master?

Things that are on the top of my head:

- Keyed children rendering - https://github.com/gopherjs/vecty/issues/25
- Optimized rendering using requestAnimationFrame, can just use https://github.com/gopherjs/vecty/issues/79
- batch rendering (I still don't follow the goals/ideas of this so we should discuss on an issue before doing a PR)
- try removing the `nil` -> `noscript` conversion, I created https://github.com/gopherjs/vecty/issues/143

Anything else you remember being in your old PR that has not yet landed in some other form?

pdf:
That's pretty much everything I think.  Batched rendering and the RAF-related code are the same thing, we can discuss later.

A quick review of my old commits did raise one thing that I haven't handled in this PR - unmounted components should be blocked from Rerender, in case an event queues a rerender while we're in the process of tearing down a tree.  I could do this by setting `Core.prevRender` to `nil` in `unmount()`, but that will trigger the panic in Rerender when this case is hit.

We could downgrade the panic to an error/warn log, or track the unmounted state via another field on `Core`.  Preference?

slimsag:
I'd like to keep the panic, so an unmounted state via another field on `Core` sounds good.

dom.go
```diff
@@ -181,7 +185,7 @@ func (h *HTML) reconcile(prev *HTML) {
 			h.node.Set(name, value)
 		}
 	}
-	if !h.new {
+	if h.node == prev.node {
```
slimsag: For readability sake I'd like to declare this as `new := h.node != prev.node` and continue using `!new` in the if statement. I'll make this change before merging.

pdf: I'm online currently, happy to edit this in if you'd like.

slimsag: If you're still online, please do :)

slimsag: Actually, just cherry-pick my commit onto your branch if you don't mind: https://github.com/gopherjs/vecty/commit/059f8593ecdd7c13ffee79da47ebdaad6448e244

dom.go
```diff
@@ -453,11 +468,18 @@ func Rerender(c Component) {
 	if prevRender == nil {
 		panic("vecty: Rerender invoked on Component that has never been rendered")
 	}
-	nextRender, skip := renderComponent(c, prevRender)
+	if c.Context().unmounted {
+		global.Get("console").Call("warn", "vecty: Rerender invoked on unmounted Component")
```
slimsag: I'm not sure I like this. I feel semi-strongly that this should either be a panic or non-existent.

In this situation, is the user of Vecty at fault / they made a programming mistake (panic)?

Or is this just something that can occur because of Rerender being incidentally invoked after a component is unmounted in an event handler (silent / no warning)?

dom.go
slimsag: Sorry, please disregard my last two comments and see my comment below instead.

slimsag:
@pdf everything here looks good except my two inline comments. My suggestion would be to use the two commits i've put into this PR: https://github.com/gopherjs/vecty/pull/145

WDYT? If those look good to you, I'll merge this one and then that one.

pdf:
Yeah, I'm fine with merging #145 on top of this.

dom.go
pdf: The user _may_ have made a programming mistake to produce this that they didn't intend, but I think it's most likely to occur due to a race in some evented code that they don't necessarily care about.

pdf:
Feels like things are starting to come together nicely FeelsGoodMan :sunglasses:

slimsag:
😄

PR #140, opened 2017-09-05
Move storeutil package to example/todomvc/store.

### LINKED ISSUE #137: Move storeutil into todomvc example subpackage
`storeutil` is not a final API design, and I feel very uncertain about multiple aspects of it. I will move it into `example/todomvc/...` and users of it today can import from there or copy if they need it.

In the future, we'll come up with a better pattern.

shurcooL:
According to https://godoc.org/github.com/gopherjs/vecty/storeutil?importers, `todomvc` is the only user of it.

Resolves #137. /cc @slimsag


slimsag:
It looks like CI is failing since it runs:

```
$ golint ./elem/... event/... prop/... storeutil/...
```

Also, we'll need to make a note in `doc/CHANGELOG.md` pointing to this PR.

shurcooL:
PTAL.

slimsag:
Looks great, thanks for sending this @shurcooL !

PR #141, opened 2017-09-05
Travis: Simplify gopherjs invocation.

shurcooL:
GopherJS supports relative import path patterns as of GopherJS 1.8-2. See gopherjs/gopherjs#644.


slimsag:
Awesome, thanks for this!

Issue #144, opened 2017-09-09
Is removing strconv feasible?


pdf:
Per https://github.com/gopherjs/gopherjs/wiki/JavaScript-Tips-and-Gotchas the `fmt` package adds significant size to the transpiled code.  Current uses in Vecty appear pretty trivial, and can likely be replaced by https://github.com/cathalgarvey/fmtless or possibly alternative calls and string concat.

slimsag:
Yeah, this is a good idea. May also be worth investigating removing our other two stdlib dependencies:

- `strconv` https://github.com/gopherjs/vecty/blob/master/style/style.go#L4
- `strings` https://github.com/gopherjs/vecty/blob/master/markup.go#L5

To start with I'd say keep it simple / don't pull in a dependency like fmtless

pdf:
I'm pretty sure `strconv` is fine, would need to check `strings` in case it pulls in `fmt`.  The wiki notes that `fmt` increases output size by ~500KB... that's a _lot_.

slimsag:
I did a quick check, one with an empty Go file just calling `println`, and three others just importing on package and calling a basic func in it.

```
$ du -sh ./*.js
 68K	./empty.js
780K	./fmt.js
108K	./strconv.js
100K	./strings.js
```

Almost certainly we can copy the relevant func that we actually use (or just use `js` package / browser API) to replace most of those easily. Our usages of them are pretty basic.

And a file importing all three:

```
792K	./fmtandstrconvandstrings.js
```

So we should be able to shave off ~700K by getting rid of those three.

Notably, those are not minified file sizes, but amount of code that must be parsed by browser is also important.

pdf:
Ack, thanks for the tests, makes things pretty clear.  I actually ended up with 939K for `fmt` on Go1.9 :cry:

slimsag:
I just retried with latest GopherJS binary & same Go version (go 1.9 darwin/amd64).

Without minification
```
$ du -sh ./*.js
 68K	./empty.js
924K	./fmt.js
936K	./fmtandstrconvandstrings.js
108K	./strconv.js
100K	./strings.js
```

With minification
```
$ du -sh ./*.js
 52K	./empty.js
624K	./fmt.js
632K	./fmtandstrconvandstrings.js
 80K	./strconv.js
 72K	./strings.js
```

slimsag:
So that's ~868K we can save without minification, and ~580K with just by ditching those dependencies.

shurcooL:
[...]

PR #146, opened 2017-09-10
Remove imports of fmt and strings packages, to reduce binary size

pdf:
Helps #144 

Looks like `strconv` is imported by something fairly standard.  I didn't try too hard to track it down, I just removed `fmt` and `strings`.

### DIFF
--- a/example/todomvc/store/storeutil/storeutil.go
+++ b/example/todomvc/store/storeutil/storeutil.go
@@ -1,8 +1,6 @@
 // Package storeutil contains a ListenerRegistry type.
 package storeutil
 
-import "fmt"
-
 // ListenerRegistry is a listener registry.
 // The zero value is unfit for use; use NewListenerRegistry to create an instance.
 type ListenerRegistry struct {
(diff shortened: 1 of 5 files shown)


example/todomvc/store/storeutil/storeutil.go
```diff
@@ -24,7 +22,7 @@ func (r *ListenerRegistry) Add(key interface{}, listener func()) {
 		key = new(int)
 	}
 	if _, ok := r.listeners[key]; ok {
-		panic(fmt.Sprintf("listener with key already exists: %v", key))
+		panic("duplicate listener key")
```
slimsag: should still include key here, otherwise the panic message is substantially less useful.

slimsag:
One comment I'd like to see addressed before merging.

I've also edited PR description to remove:

> Fixes #144

Because I'd still like to figure out if removing strconv is feasible. I'll rename the issue to be just for strconv once merged.

example/todomvc/store/storeutil/storeutil.go
pdf: key is an interface, could be anything (struct, pointer, whatever).

pdf: Also, this code is in the examples, and will probably go away at some point anyway, which is why I didn't worry too much about it.

PR #147, opened 2017-09-16
Keyed children

### LINKED ISSUE #25: Better list element reuse / "keys"

pdf:
Initial pass at keyed children.  This one was pretty tough due to having to handle keys and insertions for nested `List` children.

Working through this did show up some bugs in the existing implementation of `List` handling, primarily: the `*HTML` `nil` type coercion from #138 was not applying to `List` children, so I've moved that into `*HTML.reconcileChildren()`, as I need to handle `List`s there anyway.

Fixes #25

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -261,6 +261,16 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts 
 	for i, nextChild := range h.children {
 		// TODO(pdf): Add tests for node equality
 		new := h.node != prev.node
+
+		// Massage concrete type if necessary.
+		switch v := nextChild.(type) {
+		case *HTML:
+			if v == nil {
+				nextChild = nil
+				h.children[i] = nextChild
+			}
+		}
+
 		if i >= len(prev.children) || new {
 			if nextChildList, ok := nextChild.(List); ok {
 				nextChildList.reconcile(h.node, insertBefore, nil)
--- a/dom.go
+++ b/dom.go
@@ -68,11 +68,25 @@ type Unmounter interface {
 	Unmount()
 }
 
+// Keyer is an optional interface that a Component can implement in order to
+// uniquely identify the component amongst its siblings. If implemented, all
+// siblings, both components and HTML, must also be keyed.
+//
+// Implementing this interface allows siblings to be removed or re-ordered
+// whilst retaining state, and improving render efficiency.
+type Keyer interface {
+	// Key returns a value that uniquely identifies the component amongst its
+	// siblings. The returned type must be a valid map key, or rendering will
+	// panic.
+	Key() interface{}
+}
+
 // ComponentOrHTML represents one of:
 //
 //  Component
 //  *HTML
 //  List
+//  KeyedList
 //  nil
 //
 // If the underlying value is not one of these types, the code handling the
@@ -108,10 +122,24 @@ type HTML struct {
 	properties, attributes          map[string]interface{}
 	eventListeners                  []*EventListener
 	children                        []ComponentOrHTML
+	key                             interface{}
+	// keyedChildren stores a map of keys to children, for keyed reconciliation.
+	keyedChildren map[interface{}]ComponentOrHTML
+	// insertBeforeNode tracks the DOM node that elements should be inserted
+	// before, across List boundaries.
+	insertBeforeNode jsObject
+	// lastRendered child tracks the last child that was rendered, across List
+	// boundaries.
+	lastRenderedChild *HTML
 }
 
 // Node returns the underlying JavaScript Element or TextNode.
 func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
+
+// Key implements the Keyer interface.
+func (h *HTML) Key() interface{} {
+	return h.key
+}
 
 func (h *HTML) createNode() {
 	switch {
@@ -251,29 +279,68 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
 		h.node.Set("innerHTML", h.innerHTML)
 	}
 
-	return h.reconcileChildren(prev, nil)
+	return h.reconcileChildren(prev)
 }
 
 // reconcileChildren reconciles children of the current HTML against a previous
 // render's DOM nodes.
-func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts []Mounter) {
-	// TODO better list element reuse
+func (h *HTML) reconcileChildren(prev *HTML) (pendingMounts []Mounter) {
+	hasKeyedChildren := len(h.keyedChildren) > 0
+	prevHadKeyedChildren := len(prev.keyedChildren) > 0
 	for i, nextChild := range h.children {
-		// TODO(pdf): Add tests for node equality
-		new := h.node != prev.node
-
-		// Massage concrete type if necessary.
+		// Determine concrete type if necessary.
 		switch v := nextChild.(type) {
 		case *HTML:
+			// If the type of the child is *HTML, but its value is nil, replace
+			// the child with a concrete nil, to ensure consistent render
+			// handling.
 			if v == nil {
 				nextChild = nil
 				h.children[i] = nextChild
 			}
+		case List:
+			// Replace List with keyedList, which can handle nested keys and
+			// children.
+			nextChild = KeyedList{html: &HTML{children: v}}
+			h.children[i] = nextChild
 		}
 
-		if i >= len(prev.children) || new {
-			if nextChildList, ok := nextChild.(List); ok {
-				nextChildList.reconcile(h.node, insertBefore, nil)
+		// Ensure children implement the keyer interface consistently, and
+		// populate the keyedChildren map now.
+		//
+		// TODO(pdf): Add tests for node equality, keyed children
+		var (
+			new     = h.node != prev.node
+			nextKey interface{}
+		)
+		keyer, isKeyer := nextChild.(Keyer)
+		if hasKeyedChildren && !isKeyer {
+			panic("vecty: all siblings must have keys when using keyed elements")
+		}
+		if isKeyer {
+			nextKey = keyer.Key()
+			if hasKeyedChildren && nextKey == nil {
+				panic("vecty: all siblings must have keys when using keyed elements")
+			}
+			if nextKey != nil {
+				if h.keyedChildren == nil {
+					h.keyedChildren = make(map[interface{}]ComponentOrHTML)
+				}
+				if _, exists := h.keyedChildren[nextKey]; exists {
+					panic("vecty: duplicate sibling key")
+				}
+				// Store the keyed child.
+				h.keyedChildren[nextKey] = nextChild
+				hasKeyedChildren = true
+			}
+		}
+
+		// If this is a new element (changed type, or did not exist previously),
+		// simply add the element directly. The existence of keyed children
+		// can not be determined by children index, so skip if keyed.
+		if (i >= len(prev.children) && !hasKeyedChildren) || new {
+			if nextChildList, ok := nextChild.(KeyedList); ok {
+				nextChildList.reconcile(h, nil)
 				continue
 			}
 			nextChildRender, skip, mounters := render(nextChild, nil)
@@ -284,52 +351,108 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts 
 			if m, ok := nextChild.(Mounter); ok {
 				pendingMounts = append(pendingMounts, m)
 			}
-			if insertBefore != nil {
-				h.node.Call("insertBefore", nextChildRender.node, insertBefore.node)
-				continue
-			}
-			h.node.Call("appendChild", nextChildRender.node)
+			h.lastRenderedChild = nextChildRender
+
+			// Note: we must insertBefore not appendChild because if we're
+			// rendering inside a list with unkeyed children, we will have an
+			// insertion node here.
+			h.insertBefore(h.insertBeforeNode, nextChildRender)
 			continue
 		}
 
-		prevChild := prev.children[i]
+		var prevChild ComponentOrHTML
+		if len(prev.children) > i {
+			prevChild = prev.children[i]
+		}
+		// Find previous keyed sibling if exists, and mutate from there.
+		var stableKey bool
+		if hasKeyedChildren {
+			if prevKeyedChild, ok := prev.keyedChildren[nextKey]; ok {
+				// If the previous node rendered at this position matches the
+				// keyed previous render node, we have a stable key, and will
+				// replace later instead of inserting.
+				var prevPositionRender, prevKeyedRender *HTML
+				if _, isList := prevChild.(KeyedList); !isList {
+					prevPositionRender = extractHTML(prevChild)
+				}
+				if _, isList := prevKeyedChild.(KeyedList); !isList {
+					prevKeyedRender = extractHTML(prevKeyedChild)
+				}
+				if prevPositionRender != nil && prevKeyedRender != nil && prevPositionRender.node == prevKeyedRender.node {
+					stableKey = true
+				}
+
+				prevChild = prevKeyedChild
+				// Delete the matched key from the previous index map so that
+				// we can remove any dangling children.
+				delete(prev.keyedChildren, nextKey)
+			} else {
+				prevChild = nil
+			}
+		}
+
 		var prevChildRender *HTML
 		// If the previous child was not a list, extract the previous child
 		// render.
-		if _, isList := prevChild.(List); !isList {
+		if _, isList := prevChild.(KeyedList); !isList {
 			prevChildRender = extractHTML(prevChild)
 		}
 
-		// If the previous child render was nil, try to find the next DOM node
+		// If the previous child render was nil try to find the next DOM node
 		// in the previous render so that we can insert this child at the
 		// correct location.
-		if prevChildRender == nil {
-			for j := i + 1; j < len(prev.children) && insertBefore == nil; j++ {
-				if prevChildList, ok := prev.children[j].(List); ok {
-					insertBefore = prevChildList.firstHTML()
-				} else {
-					insertBefore = extractHTML(prev.children[j])
-				}
+		if prevChildRender == nil && h.insertBeforeNode == nil {
+			// If we have not rendered any children yet, take the insert
+			// position from the first child, if any, otherwise use the
+			// next sibling from the last rendered child.
+			if h.lastRenderedChild == nil {
+				h.insertBeforeNode = h.firstChild()
+			} else {
+				h.insertBeforeNode = h.lastRenderedChild.nextSibling()
 			}
+		}
+		// If our insertion node is the current previous child, advance to the
+		// next sibling.
+		if prevChildRender != nil && prevChildRender.node == h.insertBeforeNode {
+			h.insertBeforeNode = h.insertBeforeNode.Get("nextSibling")
 		}
 
 		// If the next child is a list, reconcile its elements in-place, and
 		// we're done.
-		if nextChildList, ok := nextChild.(List); ok {
-			nextChildList.reconcile(h.node, insertBefore, prevChild)
+		if nextChildList, ok := nextChild.(KeyedList); ok {
+			nextChildList.reconcile(h, prevChild)
 			continue
 		}
 
 		// If the previous child was a list, remove the list elements from the
 		// previous render, since we no longer have a list.
-		if prevChildList, ok := prevChild.(List); ok {
-			prevChildList.remove(h.node)
+		if prevChildList, ok := prevChild.(KeyedList); ok {
+			prevChildList.remove(h)
+			prevChild = nil
+		}
+
+		// If we're keyed and not stable, find the next DOM node from the
+		// previous render to insert before, for reordering.
+		var insertBeforeKeyedNode jsObject
+		if hasKeyedChildren && !stableKey {
+			insertBeforeKeyedNode = h.lastRenderedChild.nextSibling()
+			// If the next node is our old node, mark key as stable.
+			if prevChildRender != nil && prevChildRender.node == insertBeforeKeyedNode {
+				stableKey = true
+				insertBeforeKeyedNode = nil
+			}
 		}
 
 		// Determine the next child render.
 		nextChildRender, skip, mounters := render(nextChild, prevChild)
 		if nextChildRender != nil && prevChildRender != nil && nextChildRender == prevChildRender {
 			panic("vecty: next child render must not equal previous child render (did the child Render illegally return a stored render variable?)")
+		}
+
+		// Store the last rendered child to determine insertion target for
+		// subsequent children.
+		if nextChildRender != nil {
+			h.lastRenderedChild = nextChildRender
 		}
 
 		// If the previous and next child are components of the same type, then
@@ -340,6 +463,9 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts 
 			if nextChildComponent, ok := nextChild.(Component); ok && sameType(prevChildComponent, nextChildComponent) {
 				h.children[i] = prevChild
 				nextChild = prevChild
+				if hasKeyedChildren {
+					h.keyedChildren[nextKey] = prevChild
+				}
 			}
 		}
 		if skip {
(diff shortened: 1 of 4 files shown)


pdf:
The implementation relies on the new `Keyer` interface, as well as introducing a `vecty.Key()` `Applyer`.  Elements must use the latter, and Components must use the former.

The reason for requiring Component's to implement the interface, rather than just relying on `vecty.Key()` in the Component's return from `Render()` is that we need to access the key before we perform the render/reconciliation, so that we can obtain the correct child from the previous render.

I considered alternative approaches to avoid requiring the interface, as I recall you raising that preference last time we looked at this, but I couldn't find a reasonable approach.

dom.go
```diff
@@ -69,6 +69,13 @@ type Unmounter interface {
 	Unmount()
 }
 
+// Keyer is an optional interface that a Component can implement in order to
+// uniquely identify the component amongst its siblings. If implemented, all
+// sibling components and HTML must also be keyed.
```
slimsag: As it stands:

> If implemented, all sibling components and HTML must also be keyed.

Makes it sounds like "all sibling components" *and* "HTML" must be keyed, so maybe even children of children? I think that's not the intent, so better wording would be:

> If implemented, all siblings, both component and HTML, must also be keyed.

dom.go
```diff
@@ -69,6 +69,13 @@ type Unmounter interface {
...
+type Keyer interface {
+	Key() string
```
slimsag: Add docstring:

```
> Key returns the key value that uniquely identifies this component amongst its siblings.
```

dom.go
```diff
@@ -104,16 +111,25 @@ type RenderSkipper interface {
...
+	keyMap                               map[string]ComponentOrHTML
+	nestedKeys                           bool
+	insertBeforeNode                     jsObject
+	lastRenderedChild                    *HTML
 }
 
 // Node returns the underlying JavaScript Element or TextNode.
 func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
 
+// Key satisfies the Keyer interface.
```
slimsag: Replace `satisfies` with `implements`

dom.go
slimsag: Why is a `Keyer` using a `string` key and not an `interface{}` key? I think that cases like this will be very common, and `interface{}` equality works just the same for us internally as `string` equality would (i.e. we don't need to convert to a string).

example/todomvc/components/itemview.go
```diff
@@ -81,6 +88,7 @@ func (p *ItemView) Render() *vecty.HTML {
 				"completed": p.Item.Completed,
 				"editing":   p.editing,
 			},
+			prop.ID(p.Item.Title),
```
slimsag: What is this for ?

dom.go
pdf: It could be an `interface{}`, string was just the most sensible concrete type to me.  I'm okay with changing it though.

example/todomvc/components/itemview.go
pdf: My bad, I left this in from my testing - just made it easier to identify items from devtools, will remove.

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+		var (
+			new                   = h.node != prev.node
+			nextKey               string
+			stableKey             bool
+			insertBeforeKeyedNode jsObject
+		)
+		if keyer, ok := nextChild.(Keyer); ok {
+			nextKey = keyer.Key()
+			if len(h.keyMap) > 0 && nextKey == "" {
+				panic("vecty: all siblings must have keys when using keyed elements")
```
slimsag: I agree with the panic, but `nextKey == ""` places an unusual constraint that `""` is not a valid key. As mentioned above, I'd prefer we switch to `interface{}` key types, in which case we can use `nil` here I suppose.

pdf: Zero value needs to be invalid for whatever type is used, yes.

pdf:
Let me know when you're ready for me to do another pass on this, I may have some time one night during the week.

dom.go
slimsag: Yes, let's change it

dom.go
```diff
@@ -347,89 +465,152 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts
...
 
-// remove the List's elements from the DOM node
-func (l List) remove(node jsObject) {
-	nextHTML := &HTML{node: node}
-	nextHTML.removeChildren(&HTML{node: node, children: l})
+// remove the childList's elements from the DOM node
+func (l childList) remove(next *HTML, insertBeforeNode jsObject) {
+	l.html.node = next.node
+	l.html.insertBeforeNode = next.insertBeforeNode
+	l.html.removeChildren(l.html.children)
```
slimsag: The documentation on this function and its parameters do not make sense to me.

If the goal is to `remove the childLists's elements from the DOM node`, then why does it also take a `next` parameter?

Why is ther an unused `insertBeforeNode` parameter?

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+		// Massage concrete type if necessary.
+		switch v := nextChild.(type) {
+		case *HTML:
+			if v == nil {
+				nextChild = nil
+				h.children[i] = nextChild
+			}
+		case List:
+			nextChild = childList{html: &HTML{children: v}}
+			h.children[i] = nextChild
```
slimsag: `h.children`'s data type is `[]ComponentOrHTML` which does not include `childList`. It feels wrong to ignore this.

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
 	for i, nextChild := range h.children {
-		// TODO(pdf): Add tests for node equality
-		new := h.node != prev.node
-		if i >= len(prev.children) || new {
-			if nextChildList, ok := nextChild.(List); ok {
-				nextChildList.reconcile(h.node, insertBefore, nil)
+		// Massage concrete type if necessary.
+		switch v := nextChild.(type) {
+		case *HTML:
+			if v == nil {
```
slimsag: Document.

```
// The next child's type is *HTML but it is nil. We want the same effective
// handling as if it was a nil ComponentOrHTML, so explicitly set nextChild to
// nil now.
```

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+				nextChild = nil
+				h.children[i] = nextChild
+			}
+		case List:
+			nextChild = childList{html: &HTML{children: v}}
```
slimsag: Document. There is a reason why we are replacing `List` types with our own `childList` type, but I do not know what it is.

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+				}
+				if _, exists := h.keyMap[nextKey]; exists {
+					panic("vecty: duplicate sibling key: " + nextKey)
+				}
+				h.keyMap[nextKey] = nextChild
+			}
+		}
+		if !keyed {
+			keyed = len(h.keyMap) > 0
+		}
```
slimsag: Does `keyed` need to be persistent outside the scope of the for loop?

i.e., why don't we just declare `keyed := len(h.keyMap) > 0` before the first usage below here?

dom.go
pdf: I'll rename `next` (the target for the operation) and update the docs.  The `insertBeforeNode` was used before this commit, but now is not, will remove.

dom.go
```diff
@@ -347,89 +465,152 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts
...
-		if prevChildList, ok := prevChild.(List); ok {
+func (h *HTML) removeChildren(prevChildren []ComponentOrHTML) {
+	for _, prevChild := range prevChildren {
+		if prevChildList, ok := prevChild.(childList); ok {
 			// Previous child was a list, so remove all DOM nodes in it.
-			prevChildList.remove(h.node)
+			prevChildList.remove(h, h.insertBeforeNode)
+			if h.insertBeforeNode != nil {
+				h.insertBeforeNode = prevChildList.html.insertBeforeNode
+			}
```
slimsag: I don't understand what this if block is doing. Please document.

dom.go
pdf: ComponentOrHTML is exported, but this is an unexported type, what would you like to do?

dom.go
```diff
@@ -104,16 +111,25 @@ type RenderSkipper interface {
...
-	children                        []ComponentOrHTML
+	namespace, tag, text, innerHTML, key string
+	styles, dataset                      map[string]string
+	properties, attributes               map[string]interface{}
+	eventListeners                       []*EventListener
+	children                             []ComponentOrHTML
+	keyMap                               map[string]ComponentOrHTML
+	nestedKeys                           bool
+	insertBeforeNode                     jsObject
+	lastRenderedChild                    *HTML
```
slimsag: The fields you've added here are not self-explanatory like the ones prior to this change. `key` is self explanatory, but:

- `keyMap`
- `nestedKeys`
- `insertBeforeNode`
- `lastRenderedChild`

are not. Can you document each of them with an explanation of what they are / what state they carry?

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+		case List:
+			nextChild = childList{html: &HTML{children: v}}
+			h.children[i] = nextChild
+		}
+
+		// TODO(pdf): Add tests for node equality, keyed children
+		var (
+			new                   = h.node != prev.node
+			nextKey               string
+			stableKey             bool
```
slimsag: Why do we not define this variable closer to where it is used down below?

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+			insertBeforeKeyedNode jsObject
```
slimsag: Same with this one

dom.go
pdf: I think this is fine.

dom.go
pdf: The first line is documented at [diff](https://github.com/gopherjs/vecty/pull/147/files/1371c31a8088ea66eb45d93dba67c0f75d4aabb0#diff-cf11d578486f49277bfe20d59ea10cecR419), the last three are the same as [diff](https://github.com/gopherjs/vecty/pull/147/files/1371c31a8088ea66eb45d93dba67c0f75d4aabb0#diff-cf11d578486f49277bfe20d59ea10cecR408), I'll copy the comment.

dom.go
```diff
@@ -252,19 +268,59 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
...
+					panic("vecty: duplicate sibling key: " + nextKey)
+				}
+				h.keyMap[nextKey] = nextChild
+			}
+		}
+		if !keyed {
+			keyed = len(h.keyMap) > 0
+		}
+
+		if (i >= len(prev.children) && !keyed) || new {
```
slimsag: 1. This block is purely responsible for handling creation of new child elements that the previous DOM node's children did not have, so we should document that now.

2. Why does this condition on `!keyed` ? Why would new children that are keyed _not_ be handled here? I don't understand.

dom.go
pdf: If you'd prefer to split these up I can.

dom.go
```diff
@@ -275,46 +331,99 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts
...
-			h.node.Call("appendChild", nextChildRender.node)
+			h.insertBefore(h.insertBeforeNode, nextChildRender, nil)
 			continue
 		}
 
-		prevChild := prev.children[i]
+		var prevChild ComponentOrHTML
+		if len(prev.children) > i {
+			prevChild = prev.children[i]
+		}
```
slimsag: This change introduces a lot of uncertainty for me about whether or not at this point in the loop we're dealing with children that have a direct relationship to `prev` or not. It seems this is only the case when `keyed == true`, so I think it'd be more clear to have this `prevChild` declaration be inside the `if` below and then keep the old `prevChild` declaration that you removed but have it defined below this `keyed` if statement.

dom.go
pdf: 1. Ack
2. Because position can not be used for anything when dealing with keyed children - the feature is designed to deal with children that may move position, and that means they may appear at a higher index on the current render, than the last render had.  Once we know we're dealing with keyed children, the key is the only authoritative index.

dom.go
pdf: The `if` check here is required, because if we're dealing with keyed children, the [earlier condition](https://github.com/gopherjs/vecty/pull/147/files/1371c31a8088ea66eb45d93dba67c0f75d4aabb0#diff-cf11d578486f49277bfe20d59ea10cecR317) may not have `continue`d the children loop when we get here - we still want to assign this var in either case, it just needs this check now to avoid overflowing the prev.children array when keyed.

slimsag:
1. Excluding `string` -> `interface{}`, I like the public API and am fine with using the `Keyer` interface.
2. I'm fine with / agree with the `*HTML` `nil` change.

.. with the above said; I'm afraid of this change right now because I've spent a considerable amount of time trying to grok this but still do not fully understand the implementation. With 2-3 hours yesterday, and 2 hours just now, I'm out of time and probably won't have time to take another look until next weekend.

I feel confident you've done your due diligence, and that this is a sound change, but this currently adds an additional level of complexity that I currently cannot reason about. I'm not saying we shouldn't merge it, and also not saying we should block it on testing. But in its current state I will need more time (maybe 1 - 3 weekends) to fully parse and internalize how this is all working together.

I think the main thing I do not understand is how, after this change, the overall reconciliation flow is operating. Which conditions are handling new or existing children, keyed or not, etc. is all very overwhelming currently.

I'll come back to this when I have some more spare time, hopefully next weekend. If you'd like to take a second pass on this in the meantime, feel free to of course.

pdf:
I can understand that, `List`s in particular make things somewhat complicated.  If you think it might help, we could schedule a call on Slack or something and talk it through one weekend, let me know.

In the mean time, I may take another pass and at least address the notes here, see if there's any way to make things easier to parse.

The silver lining here though, is that I believe this is the last outstanding major architectural change, based on the current issue list.

dom.go
pdf: I'll have to try and document very clearly that keys for all siblings must be of the same type, and that the returned value's type must be comparable, maybe include a link to https://golang.org/ref/spec#Comparison_operators

pdf:
We got a little side-tracked with details, but when I pushed this I also wanted to discuss some alternative implementation details.  There's no need to understand the current implementation to discuss these.

The implementation I've pushed here requires _all_ siblings to be keyed, _including_ those that are rendered as part of a `vecty.List``.  This doesn't feel ideal to me, since one of the nice things about `vecty.List` otherwise is that it provides a means to segregate a set of children that may change in length from those around them, without requiring an additional DOM node.

The only options I could think of to allow a separate keyspace for `List` children are:
1. Disallow `List` as a sibling to keyed children, I'm not sure I like having to impose this restriction, but it might be a valid constraint, and has the lowest implementation cost (with a couple of minor caveats).
2. Try to treat `List`s separately when mixed with keyed siblings.  The problem with this is that if the `List` moves position, there's no opportunity to reconcile against the previous `List`, so we'd end up re-rendering the entire `List`'s contents.  One strategy we might employ to try and deal with this is to mark prev `List`s as processed, and do a reverse-search, then a forward-search through children to try and find a `List` to work from.  This approach is the most error-prone (as far as render-efficiency, and persistence), but I think it _feels_ best for the user... when it works.
3. Find some method to provide a key to a `List` itself.  This feels kind of weird as a user though - generally keys are going to be assigned to a set of children based on some common property derived programmatically, but the list container will not usually fit into this model.  I also don't know exactly how we'd enable setting a key on a list, if we wanted to go this way.

I'll find some time to post some examples of what I'm talking about, if this isn't clear enough.  Please let me know your thoughts @slimsag.

dom.go
norunners: In Go terminology, interfaces are satisfied not implemented. However, it's good to keep the documentation consistent for this code base.

slimsag:
> The implementation I've pushed here requires all siblings to be keyed, including those that are rendered as part of a vecty.List. This doesn't feel ideal to me, since one of the nice things about vecty.List otherwise is that it provides a means to segregate a set of children that may change in length from those around them, without requiring an additional DOM node.

Good point; thank you for flagging this.

I agree that this does not sound ideal, one reason in particular is because a `vecty.List` may be generated by a library and have keys that conflict with other siblings keys.

1. This option does not sound particularly appealing, but it may be the situation that I will become convinced after seeing the difficulty in finding a better solution.
2. The trouble with this solution is that the semantics are not clear to the user, I think? If we do this, we're essentially left with multiple sibling lists and we're trying to derive a key from their actual contents? I'd love to hear your thoughts on how we would concretely define equality for a list in this regard. Maybe we would count `HTML` tag/text types and the list with the closest match is equal? Not sure..
3. I actually think that this option sounds the most sane, aside from figuring out an API to provide the key. In this regard, a list is treated similar to e.g. creating another sibling as a `ele.Div` for example. Could you provide some examples where you think this API would be weird to interact with ? i.e. what are you talking about here?:

> This feels kind of weird as a user though - generally keys are going to be assigned to a set of children based on some common property derived programmatically, but the list container will not usually fit into this model.

pdf:
1. Level of difficulty is not necessarily the driver here, but also obviousness of behaviour, at least after the first time the user discovers this limitation.
2. Yeah, as much as I would like this approach to work, I just don't think there's any way to make this reliable/predictable enough.
3. Example below the break.

---

```go
type KeyedComponent struct {
        vecty.Core
        ID string `vecty:"prop"`
}

func (c *KeyedComponent) Key() interface{} {
        return c.ID
}

func (c *KeyedComponent) Render() *vecty.HTML {
        return elem.Div(vecty.Text(c.ID))
}

type MyComponent struct {
        vecty.Core
        data []string
}

        var contents []vecty.MarkupOrChild
        for _, v := range c.data {
                contents = append(contents, &KeyedComponent{ID: v})
        }
        contents = append(contents, vecty.List(some_footer_elements)) // <- this will panic with missing key error on the List when rendered
        return elem.Div(contents...)
}
```

So, the user has a bunch of keyed components (could just as easily be keyed elements), that are generated from some data store.  This is the most common reason to use keys - dealing with arbitrary data that the developer doesn't control, and may change between renders.  With option 3, the user is left with the requirement to add some arbitrary key they make up to the `List` sibling, and hope that it doesn't conflict with any of the keys in the generated content.

Of course, it could be argued that this is poorly written, and to avoid needing to make up a key for the footer `List`, the same thing could be written as:
```diff
 func (c *MyComponent) Render() *vecty.HTML {
-        var contents []vecty.MarkupOrChild
+        var contents vecty.List
         for _, v := range c.data {
                 contents = append(contents, &KeyedComponent{ID: v})
         }
-       contents = append(contents, vecty.List(some_footer_elements)) // <- this will panic with missing key error on the List when rendered
-       return elem.Div(contents...)
+       return elem.Div(
+               contents,
+               vecty.List(some_footer_elements),
+       )
 }
```

But in that case, option 1 works just as well, and the panic in the original implementation would be more explicit about what the problem is - `vecty.List` can not be mixed with keyed children.

dom.go
slimsag: Could we change it to an `interface{}` type without requiring extension copying / conversion all over the place? (I am not sure).

Otherwise, maybe we just need a doc comment on the definition of `children` now:

```
// important: children is actually ComponentOrHTML OR childList here, from an internal perspective.
```

slimsag:
I think I understand. I'll be brief since I don't have much time -- so weigh this comment with less weight than usual as there could be a critical piece I have not understood yet.

My thinking was that we might add something like `vecty.ListWithKey` (poor name, but not important), which would make your first example look like this:

```diff
type KeyedComponent struct {
        vecty.Core
        ID string `vecty:"prop"`
}

func (c *KeyedComponent) Key() interface{} {
        return c.ID
}

func (c *KeyedComponent) Render() *vecty.HTML {
        return elem.Div(vecty.Text(c.ID))
}

type MyComponent struct {
        vecty.Core
        data []string
}

func (c *MyComponent) Render() *vecty.HTML {
        var contents []vecty.MarkupOrChild
        for _, v := range c.data {
                contents = append(contents, &KeyedComponent{ID: v})
        }
-        contents = append(contents, vecty.List(some_footer_elements)) // <- this will panic with missing key error on the List when rendered
+        contents = append(contents, vecty.ListWithKey("listkey", some_footer_elements)) // <- this does not panic because vecty.ListWithKey was used instead of vecty.List
        return elem.Div(contents...)
}
```

The idea being that you assign a key for the entire list itself, rather than for individual elements of the list. In this case `"listkey"` just needs to be unique against siblings, i.e. the keys of whatever is in `contents` (`ID: v` in this case).

Maybe i've misunderstood something, though?

pdf:
> The idea being that you assign a key for the entire list itself, rather than for individual elements of the list. In this case "listkey" just needs to be unique against siblings, i.e. the keys of whatever is in contents (ID: v in this case).

I'm not entirely certain how to make it consistent, ie - what is the method signature of `ListWithKey` here?  I suppose something like this might be acceptable?

```go
type KeyedList struct {
    key interface{}
    list LIst
}

func ListWithKey(key interface{}, List) KeyedList {}
```

---

> Maybe i've misunderstood something, though?

My example may have been based on a lack of imagination, and being too close to the code, now that I think about it further.  I had not considered the possibility that a user may want to generate a list of keyed `List`, and so I'd assumed that the only time a `List` would be used in conjunction with other keyed children would be if it was instantiated manually, with an arbitrary key, and my example was trying to demonstrate that there's no way to safely manually key such a sibling, if the data source for the other key data is outside the user's control.  I think this was bad reasoning though, and there may be some valid use-cases that I had not considered, so I believe we should disregard this line of thinking, humble apologies for the distraction.

slimsag:
> I suppose something like this might be acceptable?

Yes, I think that a signature like that would be ideal. Other semantics which probably aren't important for validating that this idea would work:

- I will suggest the name `ListKey` now.
- `KeyedList` probably does not need to be exported. Maybe just use the interface that accepts this type. `vecty.MarkupOrChild`
- Given the name, key should probably come last.

```Go
type keyedList struct {
    list LIst
    key interface{}
}

// ListKey wraps the given List with the given key for ordering purposes.
func ListKey(list List, key interface{}) MarkupOrChild { return keyedList{list, key} }
```

I'm not super confident about the name, but it seems better than `ListWithKey` and `KeyList` to me currently (the latter implies it's a list _of_ keys or something).

I think having `ListKey` return an interface instead of a published type is useful because it pushes people to pass around a `List` instead and, for example, return `List` from APIs that produce data and wrap them with `ListKey(l, "somekey")` at the usage site instead of at the data production site.

What do you think of this approach overall? Good enough idea to take a try at implementing it?

pdf:
> I'm not super confident about the name, but it seems better than ListWithKey and KeyList to me currently (the latter implies it's a list of keys or something).

Not sure I'm in love with `ListKey` either.  What do you think about a different approach?

```go
// WithKey returns a copy of the List wrapped with the given key, for ordering purposes.
func (l List) WithKey(key interface{}) vecty.MarkupOrChild {}
```

slimsag:
I like that API/name approach a lot! That's beautiful :) Can't believe I didn't think of that myself!

dom.go
pdf: Does the difficulty in documenting this give you any pause, or do you still think that `interface{}` is the correct return here?

dom.go
slimsag: I'm happy to be proven wrong, but I've never seen `<Method> satisfies the <Interface> interface.` over `<Method> implements the <Interface> interface.` in a docstring.

https://gotools.org/net/http#transport.go-L337 seems to agree with me.

dom.go
slimsag: I am still confident that `interface{}` is a better choice than `string`

I think that saying:

```
// Key returns a value that will be compared against sibling key values for ordering purposes.
```

Is plenty enough. I don't think we need to even link to the spec about comparison operators.

If an `interface{}` value is not comparable to another `interface{}` value, nothing bad will happen (i.e. no panic) aside from the comparison becoming false / not working as the user may incorrectly expect. For example: https://play.golang.org/p/7uK4AQbER-

Anecdotally, although not exactly the same situation, the new `sync.Map` type also doesn't document requirements for keys: https://golang.org/pkg/sync/#Map.Load

dom.go
slimsag: By "I think this is fine" do you mean the current code, or what I suggested?

dom.go
pdf: We don't strictly compare different key values, just look them up, and check for conflicts.  That said, a value that is invalid as a map key will panic: https://play.golang.org/p/HjI8bscDdz.

And `sync.Map` is a rather different scenario - it's obvious that you must adhere to `map` semantics there, whereas here, the fact that the key is stored in a map internally is an implementation detail, and is not obvious to the user.

dom.go
pdf: I think the change is fine.

dom.go
slimsag: Right, I see, I had forgotten that. How about just:

```
// Key returns a value that will be used for ordering purposes. The type must be valid as
// a map key, or a panic will occur.
```

Clear and to the point. Also, we should make sure to add something like:

```
_ = map[interface{}]struct{}{key: struct{}{}}
```

At any location where we receive a map key from the user of Vecty, as this would ensure they get a stack trace that shows exactly the location where they passed the invalid key. Presumably the compiler will optimize this out completely (but I didn't check, and it doesn't matter).

pdf: > Also, we should make sure to add something like:

I don't think we bother with this, components can implement `Keyer` however they like anyway, so would be of limited utility.

slimsag: Fine for now, I'll think about it more.

dom.go
pdf: This does need to exist outside this scope, since we use it to determine how to remove children.  I've slightly changed how this works in the next push any way, since we need to track this across renders too now.

pdf:
New pass addresses previous comments, and implements the keyed list with separate keyspace, and `WithKey()` method for `List`.

slimsag:
Thanks @pdf ! I'll do my best to review this ASAP so we can land it in this weekend.

pdf:
I'm not sure it's much easier to understand now than in the last pass, so please fire any questions you have at me.  I'm not sure how much more time I'll have to work on this over the weekend, but I'll certainly try to clarify anything you need me to.

pdf:
I have a little time today if you have thoughts/queries @slimsag.

slimsag:
Hi @pdf ! Sorry for the delayed response to your above messages!

I've just spent the last ~6 hours working through your PR case by case and feel I have a very solid understanding of how everything works now. :)

I've created a separate PR to your branch at https://github.com/pdf/vecty/pull/2 which has very individual commits + good commit messages with my suggestions. Once you've had a chance to look over them and feel good about the suggestions we can squash merge to keep a clean git history.

This change is very good! I like it quite a lot! 😃

pdf:
No problem, glad you could make sense of it now.  I'll take a look over your proposed changes now.

dom.go
```diff
@@ -346,89 +481,201 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts
...
+	// Effectively become the parent (copy its scope) so that we can reconcile
+	// our children against the prev child.
+	l.html.node = parent.node
+	l.html.insertBeforeNode = parent.insertBeforeNode
+	l.html.lastRenderedChild = parent.lastRenderedChild
+
+	switch v := prevChild.(type) {
+	case keyedList:
+		pendingMounts = l.html.reconcileChildren(v.html)
+	case *HTML, Component, nil:
```
pdf: Okay with this?

slimsag: Looks great!

slimsag:
Looks like you've just merged my changes + made the change https://github.com/gopherjs/vecty/pull/147#discussion_r142023193

If there's nothing else changed (I didn't spot anything), then I'm happy to merge this as soon as you feel good to :)

dom.go
```diff
@@ -346,89 +481,201 @@ func (h *HTML) reconcileChildren(prev *HTML, insertBefore *HTML) (pendingMounts
...
-		return nextHTML.reconcileChildren(&HTML{node: node, children: c}, insertBefore)
-	default:
-		if prev == nil {
-			return nextHTML.reconcileChildren(&HTML{node: node}, insertBefore)
-		}
-		return nextHTML.reconcileChildren(&HTML{node: node, children: []ComponentOrHTML{prev}}, insertBefore)
-	}
+// WithKey wraps the List in a Keyer using the given key.
+func (l List) WithKey(key interface{}) ComponentOrHTML {
+	return keyedList{key: key, html: &HTML{children: l}}
```
pdf: My only concern with this is that MarkupOrChild/ComponentOrHTML may end up with this `keyedList` in them, which makes the documentation inaccurate, amongst other things.

slimsag: Yes that is a good point actually...

slimsag: We could publish a `type KeyedList interface{}` value + update the docs of `MarkupOrChild`/`ComponentOrHTML`?

Then nothing else has to change.

slimsag: Actually, just make `keyedList` public with no exported fields / methods

pdf: That was my inclination, too.

pdf:
I shuffled the documentation for `KeyedList` around a little, since it was aimed at internal use.  Let me know if you think the public documentation is adequate.

slimsag:
Ready to merge once you are. I will update `doc/CHANGELOG.md` to add a line that `MarkupOrChild` and `ComponentOrHTML` can both now contain `KeyedList`

pdf:
:shipit:

slimsag:
🎉

PR #152, opened 2017-10-06
renderer: Perform Rerender in batches, gated by requestAnimationFrame

### LINKED ISSUE #79: Expose a way to do work on requestAnimationFrame

pdf:
By performing Rerender in batches, the main thread is blocked for significantly less time in busy apps, leading to better responsiveness.

This change also deduplicates multiple Rerender calls for the same Component in a single batch period. This can significantly improve performance for busy apps, but carries the caveat that Components must
not rely on side-effects triggered by transitory states inside their Render functions.

A time budget is applied to rendering each batch, in an attempt to maintain a 60fps frame-rate.

Fixes #79

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -6,6 +6,9 @@ import (
 	"github.com/gopherjs/gopherjs/js"
 )
 
+// batch renderer singleton
+var batch = &batchRenderer{idx: make(map[Component]int)}
+
 // Core implements the Context method of the Component interface, and is the
 // core/central struct which all Component implementations should embed.
 type Core struct {
@@ -708,30 +711,115 @@ func Text(text string, m ...MarkupOrChild) *HTML {
 	return h
 }
 
-// Rerender causes the body of the given component (i.e. the HTML returned by
+// Rerender causes the body of the given Component (i.e. the HTML returned by
 // the Component's Render method) to be re-rendered.
 //
-// If the component has not been rendered before, Rerender panics. If the
-// component was previously unmounted, Rerender is no-op.
+// If the Component has not been rendered before, Rerender panics. If the
+// Component was previously unmounted, Rerender is no-op.
+//
+// Rerender operates efficiently by batching renders together. As a result,
+// there is no guarantee that a calls to Rerender will map 1:1 with calls to
+// the Component's Render method. For example, two calls to Rerender may
+// result in only one call to the Component's Render method.
 func Rerender(c Component) {
 	if c == nil {
 		panic("vecty: Rerender illegally called with a nil Component argument")
 	}
-	prevRender := c.Context().prevRender
-	if prevRender == nil {
+	if c.Context().prevRender == nil {
 		panic("vecty: Rerender invoked on Component that has never been rendered")
 	}
 	if c.Context().unmounted {
 		return
 	}
-	nextRender, skip, pendingMounts := renderComponent(c, prevRender)
-	if skip {
+	batch.add(c)
+}
+
+// batchRenderer handles component re-renders by queueing and deduplicating
+// them, to be rendered on the next animation frame (via requestAnimationFrame).
+type batchRenderer struct {
+	// batch contains the list of pending components to render.
+	batch []Component
+	// idx maps components to batch indexes to allow dedup, retaining order.
+	idx map[Component]int
+	// scheduled tracks whether a batch has been scheduled for processing.
+	scheduled bool
+}
+
+// add a Component to the pending batch.
+func (b *batchRenderer) add(c Component) {
+	if i, ok := b.idx[c]; ok {
+		// Shift idx for delete.
+		for j, c := range b.batch[i+1:] {
+			b.idx[c] = j - 1
+		}
+		// Delete previously queued render.
+		copy(b.batch[i:], b.batch[i+1:])
+		b.batch[len(b.batch)-1] = nil
+		b.batch = b.batch[:len(b.batch)-1]
+	}
+	// Append and index component.
+	b.batch = append(b.batch, c)
+	b.idx[c] = len(b.batch) - 1
+	// If we're not already scheduled for a render batch, request a render on
+	// the next frame.
+	if !b.scheduled {
+		b.scheduled = true
+		requestAnimationFrame(b.render)
+	}
+}
+
+// render the pending batch.
+// TODO(pdf): Add tests for time budget and multi-pass renders.
+func (b *batchRenderer) render(startTime float64) {
+	// If the batch is empty, mark as unscheduled, and stop render cycle.
+	if len(b.batch) == 0 {
+		b.scheduled = false
 		return
 	}
-	// mountUnmount to check for node replacement
-	mountUnmount(nextRender, prevRender)
-	replaceNode(nextRender.node, prevRender.node)
-	mount(pendingMounts...)
+
+	// Drain the current batch.
+	pending := b.batch
+	b.batch = nil
+	b.idx = make(map[Component]int)
+
+	// Process batch.
+	for i, c := range pending {
+		// Skip unmounted components.
+		if c.Context().unmounted {
+			continue
+		}
+
+		// Check for remaining time budget, targeting 60fps (~16ms per frame).
+		if i > 0 {
+			elapsed := global.Get("performance").Call("now").Float() - startTime
+			budgetRemaining := (1000 / 60) - elapsed
+			avgRenderTime := elapsed / float64(i)
+			// If the budget remaining is less than 2 times the average
+			// Component render time, push the remainder of the batch to the
+			// next frame.
+			if budgetRemaining < avgRenderTime*2 {
+				b.batch = pending[i:]
+				for i, c := range b.batch {
+					b.idx[c] = i
+				}
+				break
+			}
+		}
+
+		// Perform render.
+		prevRender := c.Context().prevRender
+		nextRender, skip, pendingMounts := renderComponent(c, prevRender)
+		if skip {
+			continue
+		}
+		// mountUnmount to check for node replacement.
+		mountUnmount(nextRender, prevRender)
+		replaceNode(nextRender.node, prevRender.node)
+		mount(pendingMounts...)
+	}
+
+	// Schedule next frame.
+	requestAnimationFrame(b.render)
 }
 
 // extractHTML returns the *HTML from a ComponentOrHTML.
@@ -934,9 +1022,19 @@ func unmount(e ComponentOrHTML) {
 	}
 }
 
+// requestAnimationFrame calls the native JS function of the same name.
+func requestAnimationFrame(callback func(float64)) int {
+	return global.Call("requestAnimationFrame", callback).Int()
+}
+
 // RenderBody renders the given component as the document body. The given
 // Component's Render method must return a "body" element.
 func RenderBody(body Component) {
+	// block batch until we're done
+	batch.scheduled = true
+	defer func() {
+		requestAnimationFrame(batch.render)
+	}()
 	nextRender, skip, pendingMounts := renderComponent(body, nil)
 	if skip {
 		panic("vecty: RenderBody Component.SkipRender returned true")
@@ -987,6 +1085,8 @@ type jsObject interface {
 	Call(name string, args ...interface{}) jsObject
 	String() string
 	Bool() bool
+	Int() int
+	Float() float64
 }
 
 func wrapObject(j *js.Object) jsObject {
@@ -1032,6 +1132,14 @@ func (w wrappedObject) String() string {
 }
 func (w wrappedObject) Bool() bool {
 	return w.j.Bool()
+}
+
+func (w wrappedObject) Int() int {
+	return w.j.Int()
+}
+
+func (w wrappedObject) Float() float64 {
+	return w.j.Float()
 }
 
 var isTest bool
--- a/dom_test.go
+++ b/dom_test.go
@@ -1094,6 +1094,35 @@ func TestRerender_nil(t *testing.T) {
 // TestRerender_no_prevRender tests the behavior of Rerender when there is no
 // previous render.
 func TestRerender_no_prevRender(t *testing.T) {
+	var renderCallback func(float64)
+	global = &mockObject{
+		get: map[string]jsObject{
+			"performance": &mockObject{
+				call: func(name string, args ...interface{}) jsObject {
+					if name != "now" {
+						panic(fmt.Sprintf("expected call to now, not %q", name))
+					}
+					if len(args) != 0 {
+						panic("len(args) != 0")
+					}
+					return &mockObject{floatValue: 0}
+				},
+			},
+		},
+		call: func(name string, args ...interface{}) jsObject {
+			if name != "requestAnimationFrame" {
+				panic(fmt.Sprintf("expected call to requestAnimationFrame, not %q", name))
+			}
+			if len(args) != 1 {
+				panic("len(args) != 1")
+			}
+			var ok bool
+			if renderCallback, ok = args[0].(func(float64)); !ok {
+				panic("incorrect argument to requestAnimationFrame")
+			}
+			return &mockObject{intValue: 0}
+		},
+	}
 	got := recoverStr(func() {
 		Rerender(&componentFunc{
 			render: func() *HTML {
@@ -1103,6 +1132,7 @@ func TestRerender_no_prevRender(t *testing.T) {
 				panic("expected no SkipRender call")
 			},
 		})
+		renderCallback(0)
 	})
 	want := "vecty: Rerender invoked on Component that has never been rendered"
 	if got != want {
@@ -1142,9 +1172,34 @@ func TestRerender_identical(t *testing.T) {
 			bodySet = true
 		},
 	}
+	var renderCallback func(float64)
 	global = &mockObject{
 		get: map[string]jsObject{
 			"document": document,
+			"performance": &mockObject{
+				call: func(name string, args ...interface{}) jsObject {
+					if name != "now" {
+						panic(fmt.Sprintf("expected call to now, not %q", name))
+					}
+					if len(args) != 0 {
+						panic("len(args) != 0")
+					}
+					return &mockObject{floatValue: 0}
+				},
+			},
+		},
+		call: func(name string, args ...interface{}) jsObject {
+			if name != "requestAnimationFrame" {
+				panic(fmt.Sprintf("expected call to requestAnimationFrame, not %q", name))
+			}
+			if len(args) != 1 {
+				panic("len(args) != 1")
+			}
+			var ok bool
+			if renderCallback, ok = args[0].(func(float64)); !ok {
+				panic("incorrect argument to requestAnimationFrame")
+			}
+			return &mockObject{intValue: 0}
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


pdf:
For tests to work, I need to build out `jsObject` and the associated mocks to support the required methods for the RAF polyfill, but before I head down that path, I wanted to check whether you'd prefer to bring a copy into this code-base, or add the external dependency.  We can discuss the actual meat of the change first either way.

Below are some profiles for a bit of hacked-together code that does some fairly trivial DOM manipulation with tight timing on a `Rerender` loop:

vecty-batched-render

From left-to-right:
1. Current master
2. Batched render (this PR)
3. Deduplicated batched render

I have not included the deduplicated renderer (3) here (which omits intermediary renders for a Component that has multiple `Rerender` requests in a single batch), because I'm not entirely convinced we can do that 100% safely, since users _may_ rely on side-effects from the intermediary states inside `Component.Render`.

As you can see though, just batching renders inside `requestAnimationFrame` without further modification can yield significant gains, versus fully rendering every component piecemeal as requests come in.

We may want to consider adding a time-budget (for a target frame-rate) to each batch render if we're worried about dropping frames for large batches.  I'm not sure I'm hugely concerned by this though, mostly because it would have to be a best guesstimate approach, since we can't know up-front how long any particular component render is likely to take (we could take average or max of previous component renders, but prediction accuracy would still be low).

dom.go
```diff
@@ -717,21 +720,63 @@ func Rerender(c Component) {
...
 		return
 	}
-	nextRender, skip, pendingMounts := renderComponent(c, prevRender)
-	if skip {
-		return
+	batch.add(c)
+}
+
+// batchRenderer handles component re-renders by queueing and deduplicating
+// them, to be rendered on the next animation frame (via requestAnimationFrame).
```
slimsag: Would be good to not mention `deduplicating` here since we don't perform that.

dom.go
```diff
@@ -717,21 +720,63 @@ func Rerender(c Component) {
...
+		b.scheduled = true
+		RequestAnimationFrame(b.render)
+	}
+}
+
+// render the pending batch.
+func (b *batchRenderer) render(float64) {
+	// Drain the current batch.
+	var pending []Component
+	pending, b.batch = append(pending, b.batch...), nil
```
slimsag: A better way to write this would be:

```Go
pending := b.batch
b.batch = nil
```

Because that avoids copying the entire slice.

dom.go
pdf: Whups

raf.go
```diff
@@ -0,0 +1,79 @@
+// Extracted from github.com/oskca/gopherjs-raf
```
slimsag: I think we should not pull this in. Instead, let's just use the browser's native RAF.

GopherJS already doesn't support old browsers unless they have support for typed arrays, and according to caniuse.com the browser support for typed arrays and RAF is identical with exception of Opera Mini (which [doesn't support `setTimeout` either](https://en.wikipedia.org/wiki/Opera_Mini#JavaScript_support)):

- http://caniuse.com/#feat=typedarrays
- http://caniuse.com/#feat=requestanimationframe

The main benefits of not pulling this in are:

1. Less bytes sent to the browser.
1. Our `LICENSE` file doesn't need to explicitly mention this file as an outlier.

pdf: Logic seems sound enough.  Do you want to export the RAF wrappers?

slimsag:
Overall, I think this is a great idea. Aside from my inline comments, I would be happy to merge this as-is right now.

I think that we can (and probably should) do even more here, though, but we don't need to block merging this PR on this.

---

> I'm not entirely convinced we can do that 100% safely, since users may rely on side-effects from the intermediary states inside `Component.Render` (though this is probably bad design). We might consider adding an option to enable this behaviour, with an explanation of the caveat though.

I'd like to talk about this more.

1. Exactly how would a user write code inside `Component.Render` that relies on intermediary state side effects? Could you give an example?
2. I do not like the idea of offering this as an option. It's a bit like offering the ability in the Go runtime to disable the garbage collector -- you can do it, but all bets are off about what code you import will do. i.e., this would make a weird split in the Vecty package ecosystem where some packages support this option and others do not, or simply don't mention their support for it.

My gut says that the performance gains may be worth having the feature on by default, and that a user writing code which depends on that intermediary state would be hard enough (and strange enough) that it may not be a problem. If we support it from day one by default, people will write code supporting it. If it becomes a big problem, we can always turn it off. The same is not true if we decide to turn it on in the future.

---

> We may want to consider adding a time-budget (for a target frame-rate) to each batch render if we're worried about dropping frames for large batches. I'm not sure I'm hugely concerned by this though, mostly because it would have to be a best guesstimate approach, since we can't know up-front how long any particular component render is likely to take (we could take average or max of previous component renders, but prediction accuracy would still be low).

I think we can do a lot of really fancy things here. Some we should do, and others probably not (right now, at least).

- React fibers allow the user to specify which updates are high/low/medium priority. We could offer a similar setup, but personally I think that this places too heavy a burden on the user and is not that great in practice. I may be proven wrong here in the future, but I'd say we do not need this for now.

- For the time-budget, it's true that we do not know how long a component render is likely to take (specifically, because the browser intentionally does not tell scripts this information). I think that in practice how long something takes to render is not _that important_ of a detail. We can still do quite a lot without that information.
  - For example, we can assume that all browsers will target a 60hz monitor (60 FPS).
  - We can use some estimate factor like a 50/50 split where we assume that if our DOM manipulations (measured via `window.performance.timing`) have taken 8ms then browser rendering will also take 8ms. This heuristic would serve us well I think.
  - We would then assume if component rendering (using `window.performance.timing`) has taken > 8ms, we should push all remaining updates into the next batch. This would ensure we generally have around 60 FPS, as a rule of thumb. Obviously, this will not hold true in all cases. The important bit is that we don't strictly _need_ it to. There are two situations we need to handle:
    1. The user has application updates that are well below 16ms/frame -> we don't have to do anything. This will be the case for most applications.
    2. The user has pushed some very large application updates that will greatly exceed 16ms/frame -> and they would really prefer their app to be responsive during this time. This 50/50 heuristic would ensure that this is true generally.

raf.go
slimsag: Let's not for now.

slimsag:
Also I notice that currently we `requestAnimationFrame` when `Rerender` is called, but this means that the algorithm has a best & worst case as follows:

| description | frame 1 | frame 2 | frame 3 | frame 4 |
|--|---|---|---|--|
| worst case | `requestAnimationFrame` | component updates | `requestAnimationFrame` | componentUpdates |
| best case | `requestAnimationFrame` & component updates  | same as frame 1 | same as frame 1 | same as frame 1 |

It really depends on how the browser's `requestAnimationFrame` behaves relative to which frame we are operating in. A better solution might be to pre-emptively `requestAnimationFrame` every single frame _unless_ there is a lack of user calls to `Rerender`

slimsag:
https://github.com/gopherjs/vecty/pull/152#issuecomment-334972765 may be good to do as part of this PR.

My other thoughts in https://github.com/gopherjs/vecty/pull/152#issuecomment-334972597 are better done in separate PRs.

pdf:
> Exactly how would a user write code inside Component.Render that relies on intermediary state side effects? Could you give an example?

```go
type MyComponent struct {
  vecty.Core
  Trigger bool `vecty:"prop"`
}

  if c.Trigger {
    http.PersistSomethingToBackend()
  }
  return elem.Div()
}
```

With the following values for renders:
1. `MyComponent.Trigger = false`
2. `MyComponent.Trigger = true`
3. `MyComponent.Trigger = false`
If 2 and 3 occur in a single batch, this results in the call to the backend not occurring.

> My gut says that the performance gains may be worth having the feature on by default, and that a user writing code which depends on that intermediary state would be hard enough (and strange enough) that it may not be a problem. If we support it from day one by default, people will write code supporting it. If it becomes a big problem, we can always turn it off. The same is not true if we decide to turn it on in the future.

I think maybe once we formalise the dispatch pattern, people will may be less likely to do things like this, since there'll be an established pattern for propagating changes, and dispatching actions that interact with services, but honestly it's a little hard to tell for me right now what people are likely to do.

---

> it's true that we do not know how long a component render is likely to take (specifically, because the browser intentionally does not tell scripts this information).

I should have used the term reconcile here.  I was really talking about time inside Vecty's render code, rather than paint/layout by the browser.

> We would then assume if component rendering (using window.performance.timing) has taken > 8ms, we should push all remaining updates into the next batch. 

This is probably not bad in a lot of cases, but it's not a very efficient use of our time-budget when we have a lot of rendering to do.

---

> Also I notice that currently we requestAnimationFrame when Rerender is called... It really depends on how the browser's requestAnimationFrame behaves relative to which frame we are operating in.

Yeah, I was aware of the potential (fraction of) 1 frame delay between request and render, but didn't consider it particularly problematic.

> A better solution might be to pre-emptively requestAnimationFrame every single frame unless there is a lack of user calls to Rerender.

I had considered adding a `requestAnimationFrame` at the end of `batchRenderer.render()`, so that the next frame would run immediately, but return on re-entry without requesting a frame if the batch is empty.  Is this the sort of thing you had in mind?  This gives a stable:

| Frame 1 | Frame 2 | Frame 3 |
|----------|----------|--------|
| `requestAnimationFrame` | update & `requestAnimationFrame` | update & `requestAnimationFrame` |

Provided there are updates each subsequent frame, starting from frame 1 next Rerender call, at the small cost of processing an additional frame when there's no update.

slimsag:
> I should have used the term reconcile here. I was really talking about time inside Vecty's render code, rather than paint/layout by the browser.

Understood now.

>> We would then assume if component rendering (using window.performance.timing) has taken > 8ms, we should push all remaining updates into the next batch.
>
> This is probably not bad in a lot of cases, but it's not a very efficient use of our time-budget when we have a lot of rendering to do.

Yes, that is a good point. Perhaps a better default (which we could make configurable down the line) would be a 100/0 split between DOM manipulation / reconcilliation time and browser paint/layout time. This would mean:

1. We assume we are operating within a single step of the browser's 60hz loop; i.e. our time budget is `1000/60 == 16.666ms`. Any component rendering (reconcilliation) that takes longer than this is pushed into the next batch.
2. The next frame starts `16.666ms` later anyway, so aside from the time it takes for the browser to invoke our `requestAnimationFrame` callback there is no delay and we're using 100% of the time the browser gives us, effectively.

This is almost like not having a time budget at all, except for one important thing:

1. Because we intermittently pause every 16ms, the browser can begin layout/paint of these updates without blocking on the rest. For browsers that must do this synchronously, they'll hit roughly 30 FPS (if we assume 16ms of reconciliation == 16ms of paint/layout time); and for browsers that can do this mostly async they'll hit roughly 60 FPS (if we assume the browser can concurrently layout/paint the DOM independent of the executing JavaScript).

This also makes the first implementation of this quite direct / easy: if a batch update takes > 16.66ms then push updates to the next batch.

---

> I had considered adding a requestAnimationFrame at the end of batchRenderer.render(), so that the next frame would run immediately, but return on re-entry without requesting a frame if the batch is empty. Is this the sort of thing you had in mind? This gives a stable:

Yes, exactly. :)

slimsag:
The more I think about the deduplication renderer the more I think that the 'user code' problem mentioned here is not an issue.

For example, in the same way that a user could write code in `Render` that relies on `Rerender` consistently and consecutively calling `Render`, someone could've written code like:

```Go
type MyComponent struct {
  vecty.Core
  Trigger bool `vecty:"prop"`
}

// called in response to clicking a button.
func (c *MyComponent) onClick() {
  vecty.Rerender(c)
  if c.Trigger {
    http.PersistSomethingToBackend()
  }
}

  if c.Trigger {
    c.Trigger = !c.Trigger
  }
  return elem.Div()
}
```

This code would break after this PR because it assumes that `Rerender` will call a component's `Render` method before it returns.

Instead of worrying too much about these use cases, instead what I propose is that we update the  `vecty.Rerender` documentation. This means just adding:

```diff
 // Rerender causes the body of the given component (i.e. the HTML returned by
 // the Component's Render method) to be re-rendered.
 //
 // If the component has not been rendered before, Rerender panics. If the
 // component was previously unmounted, Rerender is no-op.
+//
+// Rerender operates efficiently by batching renders together, and because of
+// this there is no guarantee that a call to Rerender will map 1:1 consecutively
+// with calls to the component's Render method. For example, two calls to
+// Rerender may result in only one call to the component's Render method,
+// and they may be in any arbitrary order.
```

I think this would make it clear enough to users that relying on `Render` for updating state would be a bad idea, and opens up the possibility for us to do some quite interesting things in terms of render performance improvements.

pdf:
> Perhaps a better default (which we could make configurable down the line) would be a 100/0 split between DOM manipulation / reconcilliation time and browser paint/layout time.

My thinking was to do something like record the average reconcile time for components during each batch, and if we have < 2*avg remaining, push the rest of the batch to the next frame.  This is likely to produce desirable results in a pretty high number of cases, and in the worst case (where the last component rendered exceeds the frame budget), behaves as you're suggesting.

---

> The more I think about the deduplication renderer the more I think that the 'user code' problem mentioned here is not an issue.

Yeah, this is why I suggested that the cases where it would cause problems would probably be due to poor design on the user's part.  I envisaged a user might do something like:

```go
type NestedActionComponent struct {
  vecty.Core
  Trigger bool `vecty:"prop"`
}

func (c NestedActionComponent) Render() *vecty.HTML {
  if c.Trigger {
    http.PersistSomethingToBackend()
  }
  return elem.Div()
}

type MyComponent struct {
  vecty.Core
  trigger bool
}

// called in response to clicking a button.
func (c *MyComponent) onClick(vecty.Event) {
  c.trigger = !c.trigger
  vecty.Rerender(c)
}

  return elem.Div(
    vecty.Markup(event.Click(c.onClick)),
    &NestedActionComponent{Trigger: c.trigger},
  )
}
```

Which is a little less clear-cut than your example, since the logic to perform the action is separated from the event that triggers it.  I don't like this code, just noting that users may try something like it, but as I noted earlier, the dispatch/store pattern should make this sort of thing less likely.

I'm not necessarily opposed to adding dedup with your proposed comment, but we should be conscious that some users may still trip over this.

slimsag:
> My thinking was to do something like record the average reconcile time for components during each batch, and if we have < 2*avg remaining, push the rest of the batch to the next frame. This is likely to produce desirable results in a pretty high number of cases, and in the worst case (where the last component rendered exceeds the frame budget entirely), behaves as your suggestion would.

Makes sense, that sounds like it would be a better implementation than what I proposed.

---

> Yeah, this is why I suggested that the cases where it would cause problems would probably be due to poor design on the user's part. I envisaged a user might do something like:

Right. (also, just to be clear, I did understand that snippet of code the first time you posted it without the parent component).

> I'm not necessarily opposed to adding dedup with your proposed comment, but we should be conscious that some users may still trip over this.

Agreed. I'd like to add the dedup as soon as one of us has time to write it. :) I think that documentation, a more clear dispatcher pattern (as you suggested), and perhaps a vecty linter tool in the future could all help in preventing users from tripping over this. I am not too worried.

pdf:
> I'd like to add the dedup as soon as one of us has time to write it. :)

As the profiling image above might suggest, I already wrote it :wink:.  It's pretty trivial, so I'll add it back for the next push.

pdf:
Latest commit adds dedup and time budgets, should address all prior feedback I believe.

dom.go
pdf: This comment is correct now :wink:

dom.go
```diff
@@ -709,29 +712,110 @@ func Text(text string, m ...MarkupOrChild) *HTML {
...
+
+	// Process batch.
+	for i, c := range pending {
+		// Skip unmounted components.
+		if c.Context().unmounted {
+			continue
+		}
+
+		// Check for remaining time budget, targeting 60fps (~16ms per frame).
+		if i > 0 {
```
slimsag: It would be more clear to me to have this logic at the end of the `for` loop instead of in an `i > 0` conditional.

dom.go
```diff
@@ -709,29 +712,110 @@ func Text(text string, m ...MarkupOrChild) *HTML {
...
+			elapsed := global.Get("performance").Call("now").Float() - startTime
+			budgetRemaining := 16 - elapsed
```
slimsag: use `(1000/60) - elapsed` here instead

dom.go
```diff
@@ -709,29 +712,110 @@ func Text(text string, m ...MarkupOrChild) *HTML {
 }
 
 // Rerender causes the body of the given component (i.e. the HTML returned by
-// the Component's Render method) to be re-rendered.
+// the component's Render method) to be re-rendered.
```
slimsag: This should be left as-is, it is intentionally written `Component` referencing the `Component` interface, same as how `Render` is capitalized here.

dom.go
```diff
@@ -709,29 +712,110 @@ func Text(text string, m ...MarkupOrChild) *HTML {
...
+				}
+				break
+			}
+		}
+
+		// Perform render.
+		prevRender := c.Context().prevRender
+		if prevRender == nil {
+			panic("vecty: Rerender invoked on Component that has never been rendered")
+		}
```
slimsag: It just occurred to me, we should leave this check in `Rerender` since we want the panic stack trace to point to user code not to this function which gets called by the browser callback.

dom.go
```diff
@@ -709,29 +712,110 @@ func Text(text string, m ...MarkupOrChild) *HTML {
...
+	scheduled bool
+}
+
+// add a Component to the pending batch.
+func (b *batchRenderer) add(c Component) {
+	if i, ok := b.idx[c]; ok {
+		// Delete previously queued render.
+		copy(b.batch[i:], b.batch[i+1:])
+		b.batch[len(b.batch)-1] = nil
+		b.batch = b.batch[:len(b.batch)-1]
```
slimsag: I don't think that this code will work as-is.

My read of this code is that you're cutting out the `i`th element. But once you've done this, every other `b.idx` value that comes after `i` becomes invalid. That is, I think you need to update each index in `b.idx` greater than `i` to be `-1` (I could be off-by-one here, so please check my logic).

slimsag:
I left a few more comments inline that are important to address before merging this.

dom.go
pdf: You're right, `idx` needs an update here.

dom.go
pdf: I only changed this for consistency - it was the only capitalisation in this comment.

slimsag: Oh got it. Maybe change it the other way (`component` -> `Component`) ?

dom.go
pdf: I wasn't aiming for accuracy, since we're trying to come in under budget anyway.  I'll make the change for clarity though.

dom.go
pdf: If we put it at the end we need a condition like `if len(pending) - i > 1`, and offset the avg denominator and indexes.  I'll move it if you'd prefer, but not sure it makes things a lot clearer.

slimsag: Oh, got it, I missed that. Leave as-is then. My 2 am brain is not very functional it seems :)

pdf: I'm not 100% today either, so I appreciate the eyes as always :+1:

pdf:
Apparently Github is not showing the latest commit in the comments section any more, but should be available in the other tabs now.  There's no rush, so feel free to take a look over this tomorrow if you'd prefer - I don't expect any major changes to this now, so I can slip any minor changes in after work tomorrow if they're required.

slimsag:
Looks perfect! Thanks @pdf !

I have one last very minor suggestion, which I'll create as a second PR and will cc you on :)

PR #153, opened 2017-10-08
hint at map allocation size when clearing index map of batchRenderer

slimsag:
In https://github.com/gopherjs/vecty/pull/152 I hinted that I had one
last suggestion, and this is it.

When we clear `b.idx`, we allocate a new map in its place. This map is
likely to have the same size as the old `b.idx`, so using this as an
allocation size hint in the call to `make` is a good idea.

cc @pdf

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -780,7 +780,7 @@ func (b *batchRenderer) render(startTime float64) {
 	// Drain the current batch.
 	pending := b.batch
 	b.batch = nil
-	b.idx = make(map[Component]int)
+	b.idx = make(map[Component]int, len(b.idx))
 
 	// Process batch.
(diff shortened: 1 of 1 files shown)


pdf:
We don't really have any idea what size it will be, depends entirely on how busy the application is?

slimsag:
Hmm, yes. My thinking was that it would help primarily when the application is consistently busy, and wouldn't hurt when the application goes from busy -> not busy because it would just be overly-large for a single frame.

But I didn't foresee the edge case where the application is very busy -> not busy at all, in which case  the map could be quite large while not being used with this approach.

Maybe best to leave as-is and let the Go runtime handle it. I'll think about this some more.

PR #154, opened 2017-10-11
fix bug where properties/etc were not applied correctly when replacing elements 

pdf:
When an element was replaced by an element of conflicting type, but
shared properties, attributes, datasets, etc, with equal values, those
properties were not applied to the new element.

Fixes https://github.com/shurcooL/Go-Package-Store/pull/88#issuecomment-335552858

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -144,6 +144,7 @@ func (h *HTML) Key() interface{} {
 	return h.key
 }
 
+// createNode creates a HTML node of the appropriate type and namespace.
 func (h *HTML) createNode() {
 	switch {
 	case h.tag != "" && h.text != "":
@@ -159,6 +160,7 @@ func (h *HTML) createNode() {
 	}
 }
 
+// reconcileText replaces the content of a text node.
 func (h *HTML) reconcileText(prev *HTML) {
 	h.node = prev.node
 
@@ -184,6 +186,24 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
 			prev = &HTML{}
 		}
 		h.createNode()
+	}
+
+	if h.node != prev.node {
+		// reconcile properties against empty prev for new nodes.
+		h.reconcileProperties(&HTML{})
+	} else {
+		h.reconcileProperties(prev)
+	}
+
+	return h.reconcileChildren(prev)
+}
+
+// reconcileProperties updates properties/attributes/etc to match the current
+// element.
+func (h *HTML) reconcileProperties(prev *HTML) {
+	// If nodes match, remove any outdated properties
+	if h.node == prev.node {
+		h.removeProperties(prev)
 	}
 
 	// Wrap event listeners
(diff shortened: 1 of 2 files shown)


dom.go
```diff
@@ -200,8 +200,13 @@ func (h *HTML) reconcile(prev *HTML) []Mounter {
 		}
 	}
 
+	new := h.node != prev.node
 	// Properties
 	for name, value := range h.properties {
+		if new {
+			h.node.Set(name, value)
+			continue
+		}
```
slimsag: Why do this here instead of below (line 214) i.e. checking `if value != oldValue` to become `if new || value != oldValue` ?

Also, doesn't this mean the Attributes, Dataset, and Styles sections below are also all invalid since they are effectively the same logic?

pdf: > Why do this here instead of below

Just slightly cheaper

> Also, doesn't this mean...

Seems likely, I'll take another pass at this tonight.

slimsag: > Just slightly cheaper

Got it; I would suggest maybe using the `if` form I suggested above purely for readability. But I am OK either way / do not feel too strongly about this.

slimsag:
Happy to merge this once you're happy with it @pdf -- I see inline you said you'd take another pass so I'll punt on merging till then.

pdf:
What do you think about this little refactor?  The new function names are not entirely accurate, but the meaning seems clear enough.

slimsag:
This refactor looks great! Thanks for sending this @pdf ! I'll merge now for https://github.com/shurcooL/Go-Package-Store/pull/88 and if there is anything you want to follow-up on we can do it in a separate PR :)

pdf:
The commit body seems to have been lost in the squash merge, I think it was valuable for context.

slimsag:
Gah, dangit. Sorry. I was 90% certain I modified it correctly.. but I must've hit the wrong key before submitting or something. It looks like I accidently double pasted.

At this point it's too late to do something like a history rewrite, if I had caught it earlier I might have.

I've done the best thing I can think of: push a phony (empty) merge commit with the actual commit description. See 8b18b17dcbd4c7e7bd8e4bc34f75441e5ee7a6db

Sorry 🙈

pdf:
Not a big deal at all, just wanted to point it out - I've not worked with Github's squash merges either.

shurcooL:
For reference, when using the squash merge option, you get the following UI that lets you edit the commit message title (box on top) and body (bigger box on bottom):

image

Whatever you write there before pressing "Confirm squash and merge" is what ends up being the squash commit message title.

A template for commit messages I think is works well is the following:

```
folder/prefix: What changed (not why). (#NNN)

Why that change was done (motivation, rationale, etc.).
Additional details or information about the change.
```

Where #NNN is the reference to the PR, e.g., #154.

shurcooL:
I can confirm this PR fixes the https://github.com/shurcooL/Go-Package-Store/pull/88#issuecomment-335552858 issue. Thank you for the quick investigation and fix!

PR #155, opened 2017-10-13
markup: Multiple Class or ClassMap usage now appends additional classes

### LINKED ISSUE #80: ClassMap appension
See https://www.reddit.com/r/golang/comments/5qjxve/qlql_qlgopherjsvectyelectron_a_gui_for_ql_to/dczvrie/?st=iygkrkdy&sh=9aa5953a for details.

pdf:
This enables component libraries and composition of elements with
classes.

BREAKING: The function `prop.Class(string)` has been removed and
replaced with `vecty.Class(...string)`.  Migrating users must use the
new function and split their classes into separate strings, rather than
a single space-separated string.

Fixes #80

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -121,6 +121,7 @@ type HTML struct {
 	node jsObject
 
 	namespace, tag, text, innerHTML string
+	classes                         map[string]struct{}
 	styles, dataset                 map[string]string
 	properties, attributes          map[string]interface{}
 	eventListeners                  []*EventListener
@@ -243,6 +244,14 @@ func (h *HTML) reconcileProperties(prev *HTML) {
 		}
 	}
 
+	// Classes
+	classList := h.node.Get("classList")
+	for name := range h.classes {
+		if _, ok := prev.classes[name]; !ok {
+			classList.Call("add", name)
+		}
+	}
+
 	// Dataset
 	dataset := h.node.Get("dataset")
 	for name, value := range h.dataset {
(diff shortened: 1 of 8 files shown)


dom.go
```diff
@@ -243,6 +244,12 @@ func (h *HTML) reconcileProperties(prev *HTML) {
 		}
 	}
 
+	// Classes
+	classList := h.node.Get("classList")
+	for name := range h.classes {
+		classList.Call("add", name)
```
slimsag: It may be good to assume the DOM API is slow than a map check, i.e. before calling `add` check `prev`:

```
if _, ok := prev.classes[name]; !ok {
    classList.Call("add", name)
}
```

But this is just a hunch / best practice in my mind. It's probably plenty fast so I am OK with this as-is if you don't wish to update.

markup.go
```diff
@@ -142,19 +142,36 @@ func Data(key, value string) Applyer {
 	})
 }
 
+// Class returns an Appyer which applies the provided classes. Subsequent
```
slimsag: typo `Appyer` -> `Applyer`

markup.go
```diff
@@ -229,33 +246,3 @@ func Namespace(uri string) Applyer {
 		h.namespace = uri
 	})
 }
-
-// join is extracted from the stdlib `strings` package
-func join(a []string, sep string) string {
```
slimsag: nice

slimsag:
Looks great to me. A few inline comments, though. I'll leave the decision of https://github.com/gopherjs/vecty/pull/155#discussion_r144690514 up to you and we'll go with whatever you decide :)

Could you add a note to https://github.com/gopherjs/vecty/blob/master/doc/CHANGELOG.md ?

dom.go
pdf: It's probably plenty fast, but I'm fine with the change as I haven't benchmarked.

pdf:
Added tests that were missing for classes, too.

Issue #143, opened 2017-09-08
Investigate / try eliminating `nil` -> `noscript` conversion


slimsag:

pdf:
My notes for when we decide to tackle this:

The slightly curly bit here is `vecty.Rerender()`, since we don't currently know where to insert the new DOM node for the case where the re-rendered component output transitions from `nil`->`*HTML` (and we'd need a separate flag to determine first-render status).  A possible solution might be to store a reference to the next DOM node, but while I would like to sort this out at some stage, it's somewhat low-priority I think.

slimsag:
I'm tagging this as a breaking change because users could rely on the fact that we output `<noscript>` tags (but hopefully they don't).

I think we don't have to solve this issue for 1.0.0, but we do need to explicitly document `Component.Render` as not strictly outputting a `<noscript>` tag as it mentions today: https://godoc.org/github.com/gopherjs/vecty#Component

PR #158, opened 2017-10-22
Change Component.Render() signature to return ComponentOrHTML

### LINKED ISSUE #148: Consider changing component Render() sig to return ComponentOrHTML?
I was looking at a declarative router implementation, and if we defined the return of `Component.Render()` as `vecty.ComponentOrHTML` instead of `*vecty.HTML`, it would enable the writing of 'wrapper', or 'higher-order' components, that enhance the components they wrap with additional functionality, without requiring unwanted DOM elements.  This could come in handy for component libraries, event handling wrappers, routers, etc, whilst allowing the wrapped component to function fully with lifecycle events, persistence, etc.

pdf:
This allows nesting of Components, allowing the creation of wrapper
components that implement lifcycle events, but do not directly render
HTML.

BREAKING: This change breaks all existing Component implementations due
to change of method signature.

Fixes #148

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -6,6 +6,31 @@ Although v1.0.0 [is not yet out](https://github.com/gopherjs/vecty/milestone/1),
 Pre-v1.0.0 Breaking Changes
 ---------------------------
 
+## Nov 4, 2017 ([PR #158](https://github.com/gopherjs/vecty/pull/158)): major breaking change
+
+All `Component`s must now have a `Render` method which returns `vecty.ComponentOrHTML` instead of the prior `*vecty.HTML` type.
+
+This change allows for higher order components (components that themselves render components), which is useful for many more advanced uses of Vecty.
+
+### Upgrading
+
+Upgrading most codebases should be trivial with a find-and-replace across all files.
+
+From your editor:
+* Find `) Render() *vecty.HTML` and replace with `) Render() vecty.ComponentOrHTML`.
+
+From the __Linux__ command line:
+```bash
+git grep -l ') Render() \*vecty.HTML' | xargs sed -i 's/) Render() \*vecty.HTML/) Render() vecty.ComponentOrHTML/g'
+```
+
+From the __Mac__ command line:
+```bash
+git grep -l ') Render() \*vecty.HTML' | xargs sed -i '' -e 's/) Render() \*vecty.HTML/) Render() vecty.ComponentOrHTML/g'
+```
+
+Obviously, you'll still need to verify that this only modifies your `Component` implementations. No other changes are needed, and no behavior change is expected for components that return `*vecty.HTML` (as the new `vecty.ComponentOrHTML` interface return type).
+
 ## Oct 14, 2017 ([PR #155](https://github.com/gopherjs/vecty/pull/155)): major breaking change
 
 The function `prop.Class(string)` has been removed and replaced with `vecty.Class(...string)`.  Migrating users must use the new function and split their classes into separate strings, rather than a single space-separated string.
--- a/dom.go
+++ b/dom.go
@@ -13,8 +13,8 @@ var batch = &batchRenderer{idx: make(map[Component]int)}
 // core/central struct which all Component implementations should embed.
 type Core struct {
 	prevRenderComponent Component
-	prevRender          *HTML
-	unmounted           bool
+	prevRender          ComponentOrHTML
+	mounted, unmounted  bool
 }
 
 // Context implements the Component interface.
@@ -29,7 +29,7 @@ func (c *Core) Context() *Core { return c }
 // 		... additional component fields (state or properties) ...
 // 	}
 //
-// 	func (c *MyComponent) Render() *vecty.HTML {
+// 	func (c *MyComponent) Render() vecty.ComponentOrHTML {
 // 		... rendering ...
 // 	}
 //
@@ -39,7 +39,7 @@ type Component interface {
 	// If Render returns nil, the component will render as nothing (in reality,
 	// a noscript tag, which has no display or action, and is compatible with
 	// Vecty's diffing algorithm).
-	Render() *HTML
+	Render() ComponentOrHTML
 
 	// Context returns the components context, which is used internally by
 	// Vecty in order to store the previous component render for diffing.
@@ -849,14 +849,12 @@ func (b *batchRenderer) render(startTime float64) {
 		}
 
 		// Perform render.
-		prevRender := c.Context().prevRender
-		nextRender, skip, pendingMounts := renderComponent(c, prevRender)
+		prevHTML := extractHTML(c.Context().prevRender)
+		nextHTML, skip, pendingMounts := renderComponent(c, c)
 		if skip {
 			continue
 		}
-		// mountUnmount to check for node replacement.
-		mountUnmount(nextRender, prevRender)
-		replaceNode(nextRender.node, prevRender.node)
+		replaceNode(nextHTML.node, prevHTML.node)
 		mount(pendingMounts...)
 	}
 
@@ -872,19 +870,20 @@ func extractHTML(e ComponentOrHTML) *HTML {
 	case *HTML:
 		return v
 	case Component:
-		return v.Context().prevRender
+		return extractHTML(v.Context().prevRender)
 	default:
 		panic("vecty: encountered invalid ComponentOrHTML " + reflect.TypeOf(e).String())
 	}
 }
 
-// sameType returns whether first and second components are of the same type
-func sameType(first, second Component) bool {
+// sameType returns whether first and second ComponentOrHTML are of the same
+// underlying type.
+func sameType(first, second ComponentOrHTML) bool {
 	return reflect.TypeOf(first) == reflect.TypeOf(second)
 }
 
-// doCopy makes a copy of the given component.
-func doCopy(c Component) Component {
+// copyComponent makes a copy of the given component.
+func copyComponent(c Component) Component {
 	if c == nil {
 		panic("vecty: cannot copy nil Component")
 	}
@@ -911,14 +910,20 @@ func doCopy(c Component) Component {
 }
 
 // copyProps copies all struct fields from src to dst that are tagged with
-// `vecty:"prop"`.
+// `vecty:"prop"`, then sets *src = *dst.
 //
-// If src and dst are different types, copyProps is no-op.
+// If src and dst are different types or non-pointers, copyProps panics.
 func copyProps(src, dst Component) {
+	if src == dst {
+		return
+	}
 	s := reflect.ValueOf(src)
 	d := reflect.ValueOf(dst)
 	if s.Type() != d.Type() {
-		return
+		panic("vecty: internal error (attempted to copy properties of incompatible structs)")
+	}
+	if s.Kind() != reflect.Ptr || d.Kind() != reflect.Ptr {
+		panic("vecty: internal error (attempted to copy properties of non-pointer)")
 	}
 	for i := 0; i < s.Elem().NumField(); i++ {
 		sf := s.Elem().Field(i)
@@ -930,6 +935,11 @@ func copyProps(src, dst Component) {
 			df.Set(sf)
 		}
 	}
+	// Effectively set *src = *dst. This allows the replacement to propagate out of
+	// the calling function, which is needed for components which themselves return
+	// components (see how renderComponent works for components that return
+	// components).
+	s.Elem().Set(d.Elem())
 }
 
 // render handles rendering the next child into HTML. If skip is returned,
@@ -944,11 +954,11 @@ func copyProps(src, dst Component) {
 // 5. nextChild == Component && prevChild == *HTML
 // 6. nextChild == Component && prevChild == nil
 //
-func render(next, prev ComponentOrHTML) (h *HTML, skip bool, pendingMounts []Mounter) {
+func render(next, prev ComponentOrHTML) (nextHTML *HTML, skip bool, pendingMounts []Mounter) {
 	switch v := next.(type) {
 	case *HTML:
 		// Cases 1, 2 and 3 above. Reconcile against the prevRender.
-		pendingMounts := v.reconcile(extractHTML(prev))
+		pendingMounts = v.reconcile(extractHTML(prev))
 		return v, false, pendingMounts
 	case Component:
 		// Cases 4, 5, and 6 above.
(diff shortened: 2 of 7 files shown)


pdf:
I've not had as much time to work on this as I would like this weekend, so this is still missing tests, but I believe this is good functionally, and I'd like to get it out there for some feedback.

dom.go
```diff
@@ -849,14 +849,12 @@ func (b *batchRenderer) render(startTime float64) {
 		}
 
 		// Perform render.
-		prevRender := c.Context().prevRender
-		nextRender, skip, pendingMounts := renderComponent(c, prevRender)
+		prevHTML := extractHTML(c.Context().prevRender)
+		nextHTML, skip, pendingMounts := renderComponent(c, c.Context().prevRender)
```
pdf: `prevHTML` and `nextHTML` are resolved recursively, so that the first `*HTML` result is returned from any depth of nested components.  The components track their children through `Core.prevRender` as normal, and bubble the `*HTML` up to the outer component.

dom.go
```diff
@@ -872,19 +870,20 @@ func extractHTML(e ComponentOrHTML) *HTML {
...
+// sameType returns whether first and second ComponentOrHTML are of the same
+// type.
+func sameType(first, second ComponentOrHTML) bool {
 	return reflect.TypeOf(first) == reflect.TypeOf(second)
 }
 
-// doCopy makes a copy of the given component.
-func doCopy(c Component) Component {
+// copyComponent makes a copy of the given component.
+func copyComponent(c Component) Component {
```
pdf: Just a minor name change to get rid of the last of the old `do*` naming.

dom.go
```diff
@@ -911,14 +910,17 @@ func doCopy(c Component) Component {
...
+// `vecty:"prop"`, then replaces src with dst.
 //
-// If src and dst are different types, copyProps is no-op.
+// If src and dst are different types or non-pointers, copyProps panics.
 func copyProps(src, dst Component) {
 	s := reflect.ValueOf(src)
 	d := reflect.ValueOf(dst)
 	if s.Type() != d.Type() {
-		return
+		panic("vecty: internal error (attempted to copy properties of incompatible structs)")
```
pdf: I changed this to panic, because if we get here, we've skipped a check somewhere, and this likely indicates a bug in vecty.

dom.go
```diff
@@ -911,14 +910,17 @@ func doCopy(c Component) Component {
...
+	}
+	if s.Kind() != reflect.Ptr || d.Kind() != reflect.Ptr {
```
pdf: I suspect that if the component was not a pointer, subsequent `.Elem()` calls would have failed with a panic anyway, and we need this here to ensure that we can perform the `Set()` operation later.

dom.go
```diff
@@ -930,6 +932,8 @@ func copyProps(src, dst Component) {
 			df.Set(sf)
 		}
 	}
+	// Overwrite src with dst
+	s.Elem().Set(d.Elem())
```
pdf: By replacing the pointer value, it allows the replacement to propagate out of the calling function, which we need for nested components.

dom.go
```diff
@@ -972,7 +976,6 @@ func renderComponent(next Component, prev ComponentOrHTML) (h *HTML, skip bool,
 		// what properties the parent has specified during SkipRender/Render
 		// below.
 		copyProps(next, prevComponent)
-		next = prevComponent
```
pdf: Handled by `copyProps` now.

dom.go
```diff
@@ -991,19 +994,40 @@ func renderComponent(next Component, prev ComponentOrHTML) (h *HTML, skip bool,
...
 	if nextRender == nil {
 		// nil renders are translated into noscript tags.
 		nextRender = Tag("noscript")
 	}
 
-	// Reconcile the actual rendered HTML.
-	pendingMounts = nextRender.reconcile(extractHTML(prev))
+	switch v := nextRender.(type) {
+	case Component:
+		nextHTML, skip, pendingMounts = renderComponent(v, prevRender)
```
pdf: We recursively render components until we hit HTML, then bubble that up to the originating component.

pdf:
Added few notes to hopefully make review a little easier, in case these areas weren't obvious.

dom.go
```diff
@@ -872,19 +870,20 @@ func extractHTML(e ComponentOrHTML) *HTML {
 	case *HTML:
 		return v
 	case Component:
-		return v.Context().prevRender
+		return extractHTML(v.Context().prevRender)
```
slimsag: Nice, I like the recursive approach here.

dom.go
```diff
@@ -872,19 +870,20 @@ func extractHTML(e ComponentOrHTML) *HTML {
...
 	default:
 		panic("vecty: encountered invalid ComponentOrHTML " + reflect.TypeOf(e).String())
 	}
 }
 
-// sameType returns whether first and second components are of the same type
-func sameType(first, second Component) bool {
+// sameType returns whether first and second ComponentOrHTML are of the same
+// type.
+func sameType(first, second ComponentOrHTML) bool {
```
slimsag: Since this now operates on `ComponentOrHTML`, it might be nice to add a new docstring below the existing one like:

```Go
// Two *HTML are always of the same type, whereas two Component have their
// concrete underlying types compared.
```

WDYT?

dom.go
slimsag: Makes sense.

dom.go
```diff
@@ -13,8 +13,8 @@ var batch = &batchRenderer{idx: make(map[Component]int)}
 // core/central struct which all Component implementations should embed.
 type Core struct {
 	prevRenderComponent Component
-	prevRender          *HTML
-	unmounted           bool
+	prevRender          ComponentOrHTML
+	mounted, unmounted  bool
```
slimsag: I don't follow why we now have two booleans for this.

Would we end up in situations where `unmounted == true && mounted == true` ? The fact that we have two booleans seems to imply that.

Why not just keep `mounted` and use `!mounted` to indicate `unmounted` ?

dom.go
```diff
@@ -1035,6 +1068,12 @@ func mount(pendingMounts ...Mounter) {
 		if mounter == nil {
 			continue
 		}
+		if c, ok := mounter.(Component); ok {
+			if c.Context().mounted {
+				continue
+			}
+			c.Context().mounted = true
```
slimsag: `unmount` below sets both `c.Context().unmounted = true` and `c.Context().mounted = false`, so we would probably need to set `unmounted = false` here too? Or alternatively address my comment about not having separate `mounted` and `unmounted` states above

dom.go
```diff
@@ -911,14 +910,17 @@ func doCopy(c Component) Component {
 }
 
 // copyProps copies all struct fields from src to dst that are tagged with
-// `vecty:"prop"`.
+// `vecty:"prop"`, then replaces src with dst.
```
slimsag: Let's reword this to be a bit more clear:

> , then effectively sets *src = *dst

dom.go
slimsag: I think your comment here would be a good inline comment for this line instead of the existing one.

> // Effectively set *src = *dst. This allows the replacement to propagate out of
> // the calling function, which is needed for components which themselves return
> // components (see how renderComponent works for components that return
> // components).

slimsag:
I left a few suggestions & questions inline. Overall, I think this is a good change.

Since this change will affect every codebase, we'll def. want to add a clear & concise docs/CHANGELOG entry here. I've come up with this (I will copy it there once this is merged):

slimsag:
## Oct 22, 2017 ([PR #158](https://github.com/gopherjs/vecty/pull/158)): major breaking change

All `Component`s must now have a `Render` method which returns `vecty.ComponentOrHTML` instead of the prior ` *vecty.HTML` type.

This change allows for higher order components (components that themselves render components), which is useful for many more advanced uses of Vecty.

### Upgrading

Upgrading most codebases should be trivial with a find-and-replace across all files.

From your editor:

- Find `) Render() *vecty.HTML` and replace with `) Render() vecty.ComponentOrHTML`.

From the **Linux** command line:

```bash
git grep -l ') Render() \*vecty.HTML' | xargs sed -i 's/) Render() \*vecty.HTML/) Render() vecty.ComponentOrHTML/g'
```

From the **Mac** command line:

```bash
git grep -l ') Render() \*vecty.HTML' | xargs sed -i '' -e 's/) Render() \*vecty.HTML/) Render() vecty.ComponentOrHTML/g'
```

Obviously, you'll still need to verify that this only modifies your `Component` implementations. No other changes are needed, and no behavior change is expected for components that return `*vecty.HTML` (as the new `vecty.ComponentOrHTML` interface return type).

slimsag:
> I've not had as much time to work on this as I would like this weekend, so this is still missing tests, but I believe this is good functionally, and I'd like to get it out there for some feedback.

FWIW my only concern would be if this change might break existing components returning `*vecty.HTML`, and I think existing tests (semi) effectively cover that + this change looks relatively straightforward. So I'm happy to merge this without tests & we can address issues for components that do return `vecty.Component` as-needed / as they arise. Completely up to you.

dom.go
pdf: The zero value makes this difficult, and users may not implement both Mounter and Unmounter interfaces.  I'll take a look and see if this is possible, but I have doubts.

dom.go
pdf: Is it sufficient just to say:

```go
// sameType returns whether first and second ComponentOrHTML are of the same
// underlying type.
```

Otherwise we can describe various typical comparisons, *HTML to *HTML, *HTML to Component, to different components.  I don't feel strongly about it either way.

pdf:
> I'm happy to merge this without tests & we can address issues for components that do return vecty.Component as-needed / as they arise.

Since I need to do one more pass, I'll try and get at least some basic tests in place as part of that.

dom.go
slimsag: Yep that's sufficient.

dom.go
pdf: Had a couple of minutes to think about this, and I don't think it can be handled by a single var.  For a Component that implements `Unmounter`, but not `Mounter`:

1. Component instantiated: `Core.mounted == false` (zero value)
2. Component rendered: `Core.mounted == false` (no change)
3. Component unmounted: `Core.mounted == false` (unmount erroneously skipped)

And it's perfectly valid for a component to implement one but not the other.

slimsag:
Why would unmounted be false in point two of your example?

At some point before point 3, the component was mounted, regardless of
whether or not the component implements the counter interface, otherwise we
don't need to fire unmount at all.

In my thinking, at point 3 Core.mounted would be true and therefore unmount
would fire. It would not be erroneously skipped.

I'm on mobile so forgive any poor typos / formatting here. :)

pdf:
> At some point before point 3, the component was mounted, regardless of
whether or not the component implements the counter interface, otherwise we
don't need to fire unmount at all.

No, we only queue components for mount if they implement `Mounter`.  We need to fire unmount every time the user implements `Unmounter`, regardless of whether they implement `Mounter` - they may want to do something only when the component is removed.

pdf:
For further context, mount and unmount also have slightly different semantics internally.  We need to track unmount regardless of whether a component implements `Unmounter`, to determine whether a component is a valid target for actions like Rerender, so that we avoid operating on a DOM node that is no longer valid.  But mount is only used to fire lifecycle events on components, we don't otherwise care about it.

The `mounted` field is only used to dedup mount events, since nested components require us to check for mounters at a slightly awkward location.

slimsag:
I see your point now: `pendingMounts` is only populated when a component implements the `Mounter` interface.

My argument was that we could do something like.. making `pendingMounts` an `[]Component` instead of `[]Mounter`. This way we internally know the component was mounted, and as such would set `Core.mounted == true`. We would only invoke their `Mount` function if the `Component` implements `Mounter`, obviously. So no external behavior changes.

Now obviously:

1. That is not as performant (means we must build that `pendingMounts` list for every component).
2. That complicates `pendingMounts` logic, which may make my point moot, since I was just trying to clarify/unify the two separate `mounted` and `unmounted` states.

I'll think about this some more..

pdf:
Any conclusions @slimsag?

slimsag:
Sorry, I forgot about this (I've been extremely busy with work this week). Thank you for the ping!

I did think about it some more, however. Basically: I think you are right to use the two separate state booleans for now.

pdf:
No probs, I'll take another pass over it this afternoon then and address comments, add a couple of tests.

pdf:
Quick update, I have something weird going on in my tests that I can't reproduce outside of them.  I kind of want to make sure it's not indicative of a real bug, but I'm out of time this weekend.

dom.go
```diff
@@ -849,14 +849,12 @@ func (b *batchRenderer) render(startTime float64) {
 		}
 
 		// Perform render.
-		prevRender := c.Context().prevRender
-		nextRender, skip, pendingMounts := renderComponent(c, prevRender)
+		prevHTML := extractHTML(c.Context().prevRender)
+		nextHTML, skip, pendingMounts := renderComponent(c, c)
```
pdf: This was the source of my bug - always should have been passing the component here instead of `prevRender`, since this is what we're actually doing.  The new code and tests exposed this bug, as `prevRender` can be a component (of the same type) now.

pdf:
Ready for review.

doc/CHANGELOG.md
```diff
@@ -6,6 +6,31 @@ Although v1.0.0 [is not yet out](https://github.com/gopherjs/vecty/milestone/1),
 Pre-v1.0.0 Breaking Changes
 ---------------------------
 
+## Nov 4, 2017 ([PR #158](https://github.com/gopherjs/vecty/pull/158)): major breaking change
```
slimsag: Thanks for adding this! 👍

dom.go
```diff
@@ -991,19 +1000,40 @@ func renderComponent(next Component, prev ComponentOrHTML) (h *HTML, skip bool,
...
+	m := mountUnmount(nextRender, prevRender)
+	if m != nil {
+		pendingMounts = append(pendingMounts, m)
+	}
 
 	// Update the context to consider this render.
 	next.Context().prevRender = nextRender
-	next.Context().prevRenderComponent = doCopy(next)
+	next.Context().prevRenderComponent = copyComponent(next)
 	next.Context().unmounted = false
```
slimsag: Should we be setting `next.Context().mounted = true` here as well?

slimsag:
Once you respond to or address https://github.com/gopherjs/vecty/pull/158#discussion_r148922801 I will merge.

dom.go
pdf: It's not mounted until the actual mount is performed, but it can't be unmounted if we're rendering it.

slimsag: Hmm, got it. I'll have to look into this some more to understand. For now I trust you on this & will merge :)

pdf:
Apologies, as I was working on #159 I realised that there is actually an undisclosed limitation with this implementation: `Component`s will panic on render when returning `List`/`KeyedList`, even though these are in fact valid `ComponentOrHTML`.

I believe I can solve this, however it will require the same work as #143, which is somewhat non-trivial.  In the mean time, do you think pushing an update to the changelog noting this limitation is sufficient?

:disappointed:

slimsag:
Oh, I see. I am actually not too worried about this given that it's really just a 'bug' in the new feature we've added. My understanding is that it should not affect existing users who return `*HTML`, so I think it's fine to just treat like any other bug and create an issue.

I've done so here https://github.com/gopherjs/vecty/issues/160

> In the mean time, do you think pushing an update to the changelog noting this limitation is sufficient?

I think the changelog only needs to mention API breaking changes for now. In the future we would do this (post 1.0), we would do this, but because we have so many breaking changes for now I think it's important to keep the signal to noise ratio low for people looking for breaking changes.

Anyone interested in this new feature will spot the bug report. :)

PR #159, opened 2017-11-04
Solidify empty interface types ComponentOrHTML and MarkupOrChild.

### LINKED ISSUE #156: Think about how ComponentOrHTML and MarkupOrChild could be made into concrete types
Today these two types are `interface{}` and easy to mix up. For example of this, see the Slack discussion [starting here](https://gophers.slack.com/archives/C0LBM1R0C/p1508292376000175).

It'd be good if we could make these types more concrete.

pdf:
This enables compile-time type checking to ensure that types actually
satisfy these interfaces.

Fixes #156

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -20,6 +20,12 @@ type Core struct {
 // Context implements the Component interface.
 func (c *Core) Context() *Core { return c }
 
+// isMarkupOrChild implements ComponentOrHTML
+func (c *Core) isMarkupOrChild() {}
+
+// isComponentOrHTML implements ComponentOrHTML
+func (c *Core) isComponentOrHTML() {}
+
 // Component represents a single visual component within an application. To
 // define a new component simply implement the Render method and embed the Core
 // struct:
@@ -44,6 +50,11 @@ type Component interface {
 	// Context returns the components context, which is used internally by
 	// Vecty in order to store the previous component render for diffing.
 	Context() *Core
+
+	// ComponentOrHTML is satisfied by vecty.Core
+	ComponentOrHTML
+	// MarkupOrChild is satisfied by vecty.Core
+	MarkupOrChild
 }
 
 // Copier is an optional interface that a Component can implement in order to
@@ -94,7 +105,9 @@ type Keyer interface {
 //
 // If the underlying value is not one of these types, the code handling the
 // value is expected to panic.
-type ComponentOrHTML interface{}
+type ComponentOrHTML interface {
+	isComponentOrHTML()
+}
 
 // RenderSkipper is an optional interface that Component's can implement in
 // order to short-circuit the reconciliation of a Component's rendered body.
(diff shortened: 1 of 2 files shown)


dom.go
```diff
@@ -20,6 +20,12 @@ type Core struct {
 // Context implements the Component interface.
 func (c *Core) Context() *Core { return c }
 
+// isMarkupOrChild implements ComponentOrHTML
```
slimsag: typo here `s/ComponentOrHTML/MarkupOrChild/g`

dom.go
```diff
@@ -94,7 +105,9 @@ type Keyer interface {
 //
 // If the underlying value is not one of these types, the code handling the
 // value is expected to panic.
```
slimsag: It's impossible for the underlying value to not be one of these types now, so we should replace this comment with one that says the compiler enforces only the above types can implement this interface.

markup.go
```diff
@@ -55,14 +55,16 @@ type Event struct {
 //
 // If the underlying value is not one of these types, the code handling the
 // value is expected to panic.
```
slimsag: ditto

markup.go
```diff
@@ -202,10 +207,10 @@ func Markup(m ...Applyer) MarkupList {
 	return MarkupList{list: m}
 }
 
-// If returns nil if cond is false, otherwise it returns the given markup.
-func If(cond bool, markup ...ComponentOrHTML) MarkupOrChild {
+// If returns nil if cond is false, otherwise it returns the given children.
+func If(cond bool, children ...ComponentOrHTML) MarkupOrChild {
```
slimsag: thanks for changing this!

dom.go
```diff
@@ -44,6 +50,11 @@ type Component interface {
 	// Context returns the components context, which is used internally by
 	// Vecty in order to store the previous component render for diffing.
 	Context() *Core
+
+	// ComponentOrHTML is satisfied by vecty.Core
+	ComponentOrHTML
+	// MarkupOrChild is satisfied by vecty.Core
+	MarkupOrChild
```
slimsag: We should instead just add the private methods here:

```Go
    isComponentOrHTML()
    isMarkupOrChild()
```

Then these will not show up in our godoc and instead it will show:

```
type Component interface {
    ...exported methods...
    // contains filtered or unexported methods
}
```

For an example of this, see https://golang.org/pkg/reflect/#Type

slimsag:
I'll make the changes I suggested inline & merge this. I think this change (and the ones I suggested inline) are quite uncontroversial.

slimsag:
Done in https://github.com/gopherjs/vecty/pull/161

pdf:
Sorry for the copy-pastas.

slimsag:
No worries! Thank you for sending this 😄

PR #161, opened 2017-11-04
Solidify empty interface types ComponentOrHTML and MarkupOrChild. 

### LINKED ISSUE #156: Think about how ComponentOrHTML and MarkupOrChild could be made into concrete types

slimsag:
This enables compile-time type checking to ensure that types actually
satisfy these interfaces.

Fixes #156

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -103,8 +103,8 @@ type Keyer interface {
 //  KeyedList
 //  nil
 //
-// If the underlying value is not one of these types, the code handling the
-// value is expected to panic.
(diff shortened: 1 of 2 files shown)


slimsag:
Supersedes https://github.com/gopherjs/vecty/pull/159

PR #162, opened 2017-11-04
Ensure ComponentOrHTML satisfies MarkupOrChild.

pdf:
This avoids some awkwardness, and API breakage for users returning
ComponentOrHTML from funcs as children in element constructors.

Refs #159, #161, #156

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -105,6 +105,7 @@ type Keyer interface {
 // underlying value must be one of these types.
 type ComponentOrHTML interface {
 	isComponentOrHTML()
+	isMarkupOrChild()
 }
 
 // RenderSkipper is an optional interface that Component's can implement in


pdf:
I missed this case in my testing yesterday.

dom.go
```diff
@@ -105,6 +105,7 @@ type Keyer interface {
 // underlying value must be one of these types.
 type ComponentOrHTML interface {
 	isComponentOrHTML()
+	isMarkupOrChild()
```
slimsag: I'm not sure that I understand this change.

If we add this method here, then `MarkupList` is also valid `ComponentOrHTML` which is invalid according to the description.

slimsag: Nevermind, I see now this would not let `MarkupList` implement `ComponentOrHTML` because it does not have an `isComponentOrHTML` method.

But I still don't quite follow this change.

pdf: `MarkupOrChild` is a super-set of `ComponentOrHTML`.  This change is required to allow something like the following:

```go
func child() vecty.ComponentOrHTML {
    return elem.Div()
}

elem.Div(
    child(),
)
```

Without it, this will fail to compile, as `ComponentOrHTML` is does not satisfy `MarkupOrChild`.

slimsag: Ohh I see it now. Thank you for the code example. I think the main reason this is confusing is because it kind of requires thinking about the problem in an inverted way. `ComponentOrHTML` implements `isMarkupOrChild` because it is a valid `MarkupOrChild` type..

I think this would be more clear if we renamed `isMarkupOrChild` and `isComponentOrHTML` to `validComponentOrHTML` and `validMarkupOrChild`.

I will merge this now since it makes sense to me now & I can see why this is quite important, indeed.

pdf: > I think the main reason this is confusing is because it kind of requires thinking about the problem in an inverted way

Yep, and I haven't had my :coffee: this morning either :wink:.

> I think this would be more clear if we renamed isMarkupOrChild and isComponentOrHTML to validComponentOrHTML and validMarkupOrChild.

I'd be fine with that, agree it might be easier to parse.

PR #166, opened 2017-11-09
doc: add projects_using_vecty.md

slimsag:
Rendered: https://github.com/gopherjs/vecty/blob/doc-projects-using-vecty/doc/projects-using-vecty.md

/cc for those whom it may concern @shurcooL @iafan @gernest @marwan-at-work :)


slimsag:
Will merge this since it's just a doc change :) If anyone has feedback I am all ears.

PR #167, opened 2017-11-09
correctly invoke Mount for components inside a List

slimsag:
I noticed that we accidently discard `pendingMounts` for any `Component` that
implements the `Mounter` interface inside a `List`. For example:

```
vecty.If(true, &SomeComponent{})
```

`SomeComponent.Mount` would never be invoked because it returns a `List` and
would have it's `pendingMounts` discarded. This fixes that, and I've verified
we do not discard `pendingMounts` by accident anywhere else.


slimsag:
cc @pdf

pdf:
This was absolutely not intended, well spotted.

pdf:
This does illustrate our need to improve testing.

slimsag:
Cool, I mostly just wanted to make sure I wasn't off base or missed something about how the code worked. :)

I filed #168 just now for some better testing, since it's been in the back of my mind for a while now but I haven't had time to get to it yet.

Issue #171, opened 2017-11-11
Writing extensible elements is harder than it should be


slimsag:
The scenario: I want to create something like the `elem` subpackage but with more sane default values than what the browser provides. For example, buttons would be styled more nicely than the browser defaults, etc. And of course, I want the caller to be able to override these 'sane defaults' they should they want to.

Basically, I want something with the same exact signature as [`elem.Button`](https://godoc.org/github.com/gopherjs/vecty/elem#Button), just with some default `vecty.Markup`. For example, it might have a default black border style, and the caller could overwrite it by calling it like this:

```Go
Button(
	vecty.Markup(
		vecty.Style("border": "99px solid gold"),
	),
)
```

---

This is possible today, but the code isn't that pretty or straightforward:

```Go
func Button(markup ...vecty.MarkupOrChild) *vecty.HTML {
	return elem.Button(
		append([]vecty.MarkupOrChild{
			vecty.Markup(
				vecty.Style("border": "1px solid black"),
			),
		}, markup...)
	)
}
```

I wrote this in my first attempt, but it doesn't compile / work:

```Go
func Button(content ...vecty.MarkupOrChild) *vecty.HTML {
	return elem.Button(
		vecty.Markup(
			vecty.Style("border": "1px solid black"),
		),
		content...,
	)
}
```

This is because `vecty.Markup` is `MarkupList` and `content` is `[]vecty.MarkupOrChild`, but `elem.Button` is a variadic `...vecty.MarkupOrChild` and variadic arguments in Go only accept parameters of the same type.

I'm not saying that we even can do anything to make the above easier to write -- but this issue is about asking the question: can we?

Related to #149

pdf:
Is this a little better?
```go
func Button(content ...vecty.MarkupOrChild) *vecty.HTML {
	return elem.Button(
		vecty.Markup(
			vecty.Style("border": "1px solid black"),
		).(vecty.MarkupOrChild),
		content...,
	)
}
```

slimsag:
Oh interesting, I did not think of that. Indeed, that's a bit better!

---

I am also wondering what drawbacks there would be to:

```Go
func Tag(tag string, children ...vecty.ComponentOrCHild) *vecty.HTML
...
func (h *HTML) WithMarkup(markup ...Applyer) *HTML
...
Tag("button", children...).WithMarkup(markup...)
```

A custom `Button` signature then becomes:

```Go
func Button(children ...vecty.ComponentOrChild) *vecty.HTML {
[...]

PR #172, opened 2017-11-12
DO NOT MERGE: Replace vecty.Markup with (*HTML).WithMarkup

slimsag:
This is an experimental change so that we can further discuss https://github.com/gopherjs/vecty/issues/171 -- I wrote this because I wanted to see what it would actually look like. I have no intention of merging this, so don't waste time actually reviewing it :)

# Primary change

`vecty.Markup` has been removed in favor of a new `(*HTML).WithMarkup` method. This code:

```Go
elem.Body(
	vecty.Markup(
		markup...
	)
	children...
)
```

Now becomes:

```Go
elem.Body(
	children...
).WithMarkup(markup...)
```

# Complete breaking change list

(you can ignore this)

- `elem.Foobar` and `Tag` arguments `m ...MarkupOrChild` have been replaced with `children ...ComponentOrChild`.
- `Markup` has been replaced with `(*HTML).WithMarkup`
- `MarkupList` is no longer exported, use `[]Applyer` instead. TODO(slimsag): this could be a mistake because it has no `Apply` method?
- `MarkupOrChild` has been removed. Use `Applyer` to represent markup, or `ComponentOrHTML` to represent children where applicable.
- `If` now returns `ComponentOrHTML` instead of `MarkupOrChild` (no change for most users).
- `func Text(text string, m ...MarkupOrChild) *HTML` is now just `func Text(text string) *HTML` since `Text` cannot have children (use `.WithMarkup` to apply markup).
- `Tag` now accepts `ComponentOrHTML` instead of `MarkupOrChild` (use `.WithMarkup to apply markup).
- The `elem` subpackage now matches the new signature of `Tag` (use `.WithMarkup to apply markup).

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -20,9 +20,6 @@ type Core struct {
 // Context implements the Component interface.
 func (c *Core) Context() *Core { return c }
 
-// isMarkupOrChild implements MarkupOrChild
-func (c *Core) isMarkupOrChild() {}
-
 // isComponentOrHTML implements ComponentOrHTML
 func (c *Core) isComponentOrHTML() {}
 
@@ -52,7 +49,6 @@ type Component interface {
 	Context() *Core
 
 	isComponentOrHTML()
-	isMarkupOrChild()
 }
 
 // Copier is an optional interface that a Component can implement in order to
@@ -105,7 +101,6 @@ type Keyer interface {
 // underlying value must be one of these types.
 type ComponentOrHTML interface {
 	isComponentOrHTML()
-	isMarkupOrChild()
 }
 
 // RenderSkipper is an optional interface that Component's can implement in
@@ -152,13 +147,22 @@ type HTML struct {
 // Node returns the underlying JavaScript Element or TextNode.
 func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
 
+// WithMarkup returns a copy of this HTML with the given markup applied.
+func (h *HTML) WithMarkup(markup ...Applyer) *HTML {
+	cpy := *h
+	for _, m := range markup {
+		if m == nil {
+			continue
+		}
+		m.Apply(&cpy)
+	}
+	return &cpy
+}
+
 // Key implements the Keyer interface.
 func (h *HTML) Key() interface{} {
 	return h.key
 }
-
-// isMarkupOrChild implements MarkupOrChild
-func (h *HTML) isMarkupOrChild() {}
 
 // isComponentOrHTML implements ComponentOrHTML
 func (h *HTML) isComponentOrHTML() {}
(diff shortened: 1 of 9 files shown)


slimsag:
@pdf I took a quick stab at implementing what we were talking about in https://github.com/gopherjs/vecty/issues/171

- I ignored tests, just making sure the markdown and todomvc applications worked.
- I was mostly curious to see what the API felt like in the real world (with todomvc / markdown examples).
- I didn't take a stab at `WithChildren` but I assume it would be the same way / should work fine.

Overall I actually don't like this approach. It feels stranger than `vecty.Markdown` today, mostly because markdown ends up at the bottom of the `elem.Foo().WithMarkup()` call, but is still nested inside multiple calls.

For example, compare the readability of `renderHeader` here:

Before:

https://github.com/gopherjs/vecty/blob/1ba2b77c9d1fbf3503d3ddaa1cfc25dff4516ea3/example/todomvc/components/pageview.go#L74-L100

After:

https://github.com/gopherjs/vecty/blob/3fd2fe60e22110282cf7a3a45ce6921f7f0f1645/example/todomvc/components/pageview.go#L70-L90

So I personally consider this not a great option. It seems to make the code harder to read (for me, at least). Curious to hear what you think about it from a public API perspective!

pdf:
Does this read any better?

```go
func (p *PageView) renderHeader() *vecty.HTML {
        return elem.Header().Markup(
                vecty.Class("header"),
        ).Children(
                elem.Heading1().Children(
                        vecty.Text("todos"),
                ),
                elem.Form.Markup(
                        style.Margin(style.Px(0)),
                        event.Submit(p.onAdd).PreventDefault(),
                ).Children(
                        elem.Input().Markup(
                                vecty.Class("new-todo"),
                                prop.Placeholder("What needs to be done?"),
                                prop.Autofocus(true),
                                prop.Value(p.newItemTitle),
                                event.Input(p.onNewItemTitleInput),
                        ),
                ),
        )
}
```

Or for that matter:

```go
func (p *PageView) renderHeader() *vecty.HTML {
        return elem.Header.Markup(
                vecty.Class("header"),
        ).Children(
                elem.Heading1.Children(
                        vecty.Text("todos"),
                ),
                elem.Form.Markup(
                        style.Margin(style.Px(0)),
                        event.Submit(p.onAdd).PreventDefault(),
                ).Children(
                        elem.Input.Markup(
                                vecty.Class("new-todo"),
                                prop.Placeholder("What needs to be done?"),
                                prop.Autofocus(true),
                                prop.Value(p.newItemTitle),
                                event.Input(p.onNewItemTitleInput),
                        ),
                ),
        )
}
```

slimsag:
I think it still looks/reads too verbose in that case, but def. better because markdown is at the front, yeah.

I'm reconsidering what our handling of nil looks like

PR #175, opened 2017-11-21
produce better error message when vecty.Class fails

marwan-at-work:
While I was updating my repos to fix vecty breaking changes, I forgot to separate some spaced-classes in `vecty.Class` and the error message that I got was super confusing. I can imagine newcomers to be terrified and/or turned off by confusing dom error messages (picture below).

This pr produces a better error message with instructions on how to fix their code.  

**before** 

screen shot 2017-11-21 at 3 06 47 pm

**after** 

screen shot 2017-11-21 at 3 58 27 pm

### DIFF
--- a/markup.go
+++ b/markup.go
@@ -2,6 +2,7 @@ package vecty
 
 import (
 	"reflect"
+	"strings"
 
 	"github.com/gopherjs/gopherjs/js"
 )
@@ -148,8 +149,9 @@ func Data(key, value string) Applyer {
 
 // Class returns an Applyer which applies the provided classes. Subsequent
 // calls to this function will append additional classes. To toggle classes,
-// use ClassMap instead.
+// use ClassMap instead. Each class name must be passed as a separate argument.
 func Class(class ...string) Applyer {
+	mustValidateClassNames(class)
 	return markupFunc(func(h *HTML) {
 		if h.classes == nil {
 			h.classes = make(map[string]struct{})
@@ -158,6 +160,35 @@ func Class(class ...string) Applyer {
 			h.classes[name] = struct{}{}
 		}
 	})
+}
+
+// mustValidateClassNames ensures no class names have spaces
+// and panics with clear instructions on how to fix this user error.
+func mustValidateClassNames(class []string) {
+	for _, name := range class {
+		if strings.Contains(name, " ") {
+			failClassName(name)
+		}
+	}
+}
+
+// failClassName produces a helpful error message.
+// to fix the spaces in vecty.Class for example:
+// vecty.Class("hello there") will result in the following panic:
+// Spaces are not allowed in class names. Use `vecty.Class("hello", "there")` instead of `vecty.Class("hello there")`
+func failClassName(name string) {
+	names := []string{}
+	// avoids using fmt.Sprintf for bundle size optimization.
+	for _, cn := range strings.Split(name, " ") {
+		cn = "\"" + cn + "\""
+		names = append(names, cn)
+	}
+	correct := strings.Join(names, ", ")
+	incorrect := "\"" + name + "\""
+	panic(
+		"Spaces are not allowed in class names. Use `vecty.Class(" + correct + ")` " +
+			"instead of `vecty.Class(" + incorrect + ")`",
+	)
 }
 
 // ClassMap is markup that specifies classes to be applied to an element if
--- a/markup.go
+++ b/markup.go
@@ -166,10 +166,7 @@ func Class(class ...string) Applyer {
 func mustValidateClassNames(class []string) {
 	for _, name := range class {
 		if containsSpace(name) {
-			panic(
-				`vecty: invalid class string "` + name + `" with spaces passed to ` + "`vecty.Class` -- " +
-					`you must pass each name as a separate argument`,
-			)
+			panic(`vecty: invalid argument to vecty.Class "` + name + `" (string may not contain spaces)`)
 		}
 	}
 }
--- a/markup.go
+++ b/markup.go
@@ -166,7 +166,10 @@ func Class(class ...string) Applyer {
 func mustValidateClassNames(class []string) {
 	for _, name := range class {
 		if containsSpace(name) {
-			failClassName(name)
+			panic(
+				`vecty: invalid class string "` + name + `" with spaces passed to ` + "`vecty.Class` -- " +
+					`you must pass each name as a separate argument`,
+			)
 		}
 	}
 }
@@ -179,14 +182,6 @@ func containsSpace(s string) bool {
 		}
 	}
 	return false
-}
-
-// failClassName produces a helpful error message to fix the spaces in vecty.Class.
-func failClassName(name string) {
-	panic(
-		`vecty: invalid class string "` + name + `" with spaces passed to ` + "`vecty.Class` -- " +
-			`you must pass each name as a separate argument`,
-	)
 }
 
 // ClassMap is markup that specifies classes to be applied to an element if
--- a/markup.go
+++ b/markup.go
@@ -2,7 +2,6 @@ package vecty
 
 import (
 	"reflect"
-	"strings"
 
 	"github.com/gopherjs/gopherjs/js"
 )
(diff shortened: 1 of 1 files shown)


shurcooL:
What are the performance implications of doing such validation?

Is it going to be worth it to consider a "production mode", after you know the code is correct, it'd skip unnecessary error checks at run-time? OpenGL/Vulkan has something like this.

It's very likely it's way too soon to worry about something like that, but I wanted to throw the thought out there.

marwan-at-work:
@shurcooL this computation happens before vecty renders/rerenders so everything is still in the javascript process world. However, this does mean that the render functions have to do a bunch of loops inside the javascript world so you have a point there. I don't think that's a performance cost given that vecty does far more than just looping in the javascript runtime -- i.e. dom-diffing.

The idea of having a production mode build is, however, pretty interesting because there are numerous optimizations vecty and gopherjs can make. I'm open to suggestions :)

shurcooL:
My scope of the production mode I had in mind was no more than to disable unneeded extra validation code that runs in order to give better error messages, nothing more. And I only think it's worthwhile if said validation code adds a very significant performance penalty. Otherwise, the simplicity of not having modes is a better tradeoff.

slimsag:
Thanks for taking the time to write & contribute this @marwan-at-work ! Please note I may not have time to re-review this soon, and if so, I apologize ahead of time.

1. I highly agree you should've never gotten that DOM error / should've gotten a more clear/concise vecty panic.
2. I do not think that we need a production mode, until someone can say for sure there is a significant perf penalty to their code. The check here is just a `strings.Contains(name, " ")` for valid code, which I highly doubt adds any significant performance overhead. The overhead of adding a dependency such as `strings` may be one valid argument for a production mode, however.

markup.go
```diff
@@ -160,6 +162,35 @@ func Class(class ...string) Applyer {
...
+			failClassName(name)
+		}
+	}
+}
+
+// failClassName produces a helpful error message.
+// to fix the spaces in vecty.Class for example:
+// vecty.Class("hello there") will result in the following panic:
+// Spaces are not allowed in class names. Use `vecty.Class("hello", "there")` instead of `vecty.Class("hello there")`
+func failClassName(name string) {
```
slimsag: I don't think that we should take on the task of generating panics that instruct the user how to fix their code. I think just a panic saying:

```
panic: vecty: invalid class string with spaces passed to vecty.Class
```

Is enough to guide the user in the right direction. If needed, we can offer a help/documentation page for panic messages that goes into further depth about how to fix them.

pdf: Agree, though I'd append the invalid class string to that message.

markup.go
```diff
@@ -160,6 +162,35 @@ func Class(class ...string) Applyer {
 	})
 }
 
+// mustValidateClassNames ensures no class names have spaces
+// and panics with clear instructions on how to fix this user error.
+func mustValidateClassNames(class []string) {
+	for _, name := range class {
+		if strings.Contains(name, " ") {
```
slimsag: It would be nice to not have to pull in the `strings` dependency. Could you replace this with: 

```
for _, name := range class {
    for _, c := range name {
        if c == ' ' {
            panic("vecty: invalid class string with spaces passed to vecty.Class")
        }
    }
}
```

instead?

shurcooL: That leads to pretty hard to read code and questions like "why not simplify this by using `strings.Contains`" for readers down the line.

What about creating a helper func and documenting it:

```Go
// contains reports whether substr is within s.
// It's a copy of strings.Contains to avoid importing all of strings for size reasons.
func contains(s, substr string) bool { ... }
```

slimsag: Note that we do not need full substring containment here; i.e. we're more asking here "does the string contain the byte ' '"? rather than an entire substring, which is more complex to handle.

So we could write that in the form of a helper function:

```Go
// contains tells if the byte b is in the given string s.
//
// It's equivalent to strings.Contains(s, []byte(substr)[0]) but has a simpler
// implementation and doesn't require importing the entire strings package
// (for size reasons).
func contains(s string, b byte) bool {
    for _, c := range s {
        if c == b {
            return true
        }
    }
    return false
}
```

Our other code then becomes:

```Go
for _, name := range class {
    if contains(name, ' ') {
        panic(`vecty: invalid class string "` + name + `" with spaces passed to vecty.Class`)
    }
}
```

But I'm not really sure that is any more clear than just adding a comment:

```diff
for _, name := range class {
+   // Note: we use a for loop here instead of strings.Contains to avoid pulling in the
+   // entire strings package for size reasons.
    for _, c := range name {
        if c == ' ' {
            panic(`vecty: invalid class string "` + name + `" with spaces passed to vecty.Class`)
        }
    }
}
```

WDYT @shurcooL ? which of these is better? I'm not really sure

pdf: My preference would be to extract it to a function.

shurcooL: @slimsag Sorry, I made a mistake and wrote the wrong thing halfway through my reply. What I meant to suggest was:

```Go
// containsSpace reports whether s contains a space character.
func containsSpace(s string) bool {
    for _, c := range []byte(name) {
        if c == ' ' {
            return true
        }
    }
    return false
}
```

I didn't mean to suggest reimplementing all of `strings.Contains`, that'd be silly and unneeded.

marwan-at-work: i'm happy either way. pending @shurcooL 's opinion and i'll update my pr.

thanks for the quick response everyone :v:

marwan-at-work: @shurcooL ops, did not refresh my page before i submitted the comment lol. I like the `containsSpace` function. Although i don't think we need to convert name into a byte slice, since ranging over a string returns a slice of runs and `' '` is a rune character. i'll add that function :)

shurcooL: (I chose to range over bytes rather than runes simply because I figured it might be faster (it may not actually be), but you're welcome to range over runes if you prefer that.)

slimsag: Makes sense -- thanks for the insight @shurcooL !

I think ranging over bytes makes more sense, but I'm OK with either way.

marwan-at-work: @slimsag I compared the compiled versions of both functions and the []byte one does an extra copy and the bundle size is 20 lines more. Unnecessarily nitty, I know haha.

slimsag: Awesome, thanks for doing that!

markup.go
```diff
@@ -160,6 +161,34 @@ func Class(class ...string) Applyer {
 	})
 }
 
+// mustValidateClassNames ensures no class names have spaces
+// and panics with clear instructions on how to fix this user error.
+func mustValidateClassNames(class []string) {
+	for _, name := range class {
+		if containsSpace(name) {
+			failClassName(name)
```
slimsag: Remove `failClassName` and just do:

```Go
panic(`vecty: invalid argument to vecty.Class "` + name + `" (string may not contain spaces)`)
```

?

marwan-at-work: @slimsag makes sense.

markup.go
```diff
@@ -166,7 +166,10 @@ func Class(class ...string) Applyer {
 func mustValidateClassNames(class []string) {
 	for _, name := range class {
 		if containsSpace(name) {
-			failClassName(name)
+			panic(
+				`vecty: invalid class string "` + name + `" with spaces passed to ` + "`vecty.Class` -- " +
+					`you must pass each name as a separate argument`,
+			)
```
slimsag: Sorry @marwan-at-work I should've been more clear. I actually want this panic message to just state:

1. The argument to `vecty.Class` is invalid.
2. The string argument cannot contain a space.

And nothing more. That is why I suggested just writing:

```Go
panic(`vecty: invalid argument to vecty.Class "` + name + `" (string may not contain spaces)`)
```

here instead of the `failClassName` call. I didn't mean to just inline it, but rather to _also_ make the panic message more brief.

I don't think that we should advise the user on what to do (`you must pass each name as a separate argument`) but rather just tell them what they did wrong. This is more consistent with other panics inside Vecty, where it is highly impossible for us to tell the user what they should change in their code. So I think it's bad to tell the user how to fix their code here.

pdf: If we decide that more context on how to resolve various vecty errors is required, we can implement error codes or something and maintain documentation in the wiki.

marwan-at-work: @slimsag not telling them how to fix their code, but telling them exactly what is wrong with their code makes a lot of sense :)

slimsag:
Thanks @marwan-at-work for taking the time to fix this! :)

Feel free to send a PR to add yourself to https://github.com/gopherjs/vecty/blob/master/CONTRIBUTORS if you'd like :)

PR #176, opened 2017-11-22
example: remove usage of the Restorer interface

### LINKED ISSUE #174: example/todomvc using outdated Restorer API
https://github.com/gopherjs/vecty/blob/1ba2b77c9d1fbf3503d3ddaa1cfc25dff4516ea3/example/todomvc/components/itemview.go#L26-L32

We should remove this (also one in `pageview.go`).

tbruyelle:
Fix #174

This interface has been removed in PR #130, it shouldn't be used
anymore.

### DIFF
--- a/example/todomvc/components/itemview.go
+++ b/example/todomvc/components/itemview.go
@@ -23,14 +23,6 @@ type ItemView struct {
 	input     *vecty.HTML
 }
 
-// Restore implements the vecty.Restorer interface.
-func (p *ItemView) Restore(prev vecty.Component) {
-	if old, ok := prev.(*ItemView); ok {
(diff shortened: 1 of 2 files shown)


pdf:
The test failure is clearly not due to this PR, looks like megacheck was updated, and we're not pinning the version used in CI.

slimsag:
> we're not pinning the version used in CI.

I would argue that is a good thing @pdf since it found a real issue I think 😃

slimsag:
Feel free to send a PR to add yourself to https://github.com/gopherjs/vecty/blob/master/CONTRIBUTORS if you'd like @tbruyelle :)

Issue #168, opened 2017-11-09
Achieve 100% test coverage


slimsag:
Will need to make our testing suite much less verbose for this. Ideally we will just perform some Vecty operations in the test suite (e.g. `RenderBody`) and get an output list of strings or something non-verbose of exactly which DOM operations occurred.

slimsag:
@pdf FYI I haven't finished yet but I do have a quite extensive change to move us over to a nicer test suite that is more data-driven and just produces a list of expected vs. gotten DOM operations & will make it much easier to test correct behavior in Vecty without writing crazy verbose DOM mocking code. Just wanted to let you know so you don't do anything too extensive in the test suite any time soon :)

slimsag:
Actually I had some spare hours, so I moved the entire test suite over and sent https://github.com/gopherjs/vecty/pull/178 :)

Still todo here is adding more tests to achieve 100% coverage, but it should be a lot easier once that PR is merged.

PR #178, opened 2017-12-04
add new data-driven test suite + move all existing tests to it

slimsag:
This change moves us over to a data-driven test suite and moves all existing tests over to it.

The basic idea is that instead of writing a bunch of DOM mocking code like we used to, we now just create a test by performing some Vecty actions and the DOM operations are recorded to a file. We keep the file of wanted DOM operations around, and they are compared.

For example, `TestSetTitle` used to look like this:

```Go
func TestSetTitle(t *testing.T) {
	titleSet := ""
	document := &mockObject{
		set: func(key string, value interface{}) {
			if key != "title" {
				panic(fmt.Sprintf(`expected document.set "title", not %q`, key))
			}
			titleSet = value.(string)
		},
	}
	global = &mockObject{
		get: map[string]jsObject{
			"document": document,
		},
	}
	want := "foobar"
	SetTitle(want)
	if titleSet != want {
		t.Fatalf("titleSet is %q, want %q", titleSet, want)
	}
}
```

The vast majority of this code is setting up a fake JS DOM `global`, `global.get`, `document`, `document.set`, etc. objects & methods in order to properly check `was global.document.title set to the string "foobar"?`. It's far too verbose and unreadable for such a simple test. The problem is even worse for larger tests that we have.

Instead of the above, the new test suite makes this test be written just as:

```Go
// TestSetTitle tests that the SetTitle function performs the correct DOM
// operations.
func TestSetTitle(t *testing.T) {
	ts := testSuite(t, "TestSetTitle")
	defer ts.done()

	SetTitle("foobartitle")
}
```

When `go test` is ran, a new file `testdata/TestSetTitle.got.txt` is written which looks like this:

```
global.Get("document")
global.Get("document").Set("title", "foobartitle")
```

It tells us that first `global.Get("document")` was called, and subsequently `Set("title", "foobartitle")` was called on the pointer returned by `global.Get("document")` originally (that is, the second line does not say that `global.Get("document")` was called twice).

The test suite also tells you that `testdata/TestSetTitle.got.txt` does not match (the non-existent) `testdata/TestSetTitle.want.txt` file and it shows you a nice colored diff and informs you:

```bash
    	testsuite_test.go:211: to accept these changes:
    		
    		$ mv testdata/TestSetTitle.got.txt testdata/TestSetTitle.want.txt
```

If you then run the suggested command and commit the `testdata` folder changes, the test now succeeds & no regressions can occur here in the future.

Additionally, what the diff stats on GitHub here does not show is that:

1. 1,668 lines of hand-written Go test code have been removed(!)
2. Only 588 lines of Go code have been added to build this new test suite.
2. 493 lines of automatically generated `testdata/**.want.txt` files have been added.

I think that this new test suite will substantially help in our goal to achieve 100% test coverage (#168) because adding a new test requires no more mocking. Just perform some Vecty actions in the Go test code, run `go test` from the command line, confirm the DOM operations look correct &  follow the instructions when prompted. 😃 

Note: I will merge this change not as a squash commit, because it is in logical commits & is useful to see how each individual test was migrated over to the new test suite.

Helps #168

/cc @pdf (and @shurcooL who may find this to be an interesting/cool simplification :) )


gedw99:
This is awesome.

slimsag:
This change looks generally well received (two hearts + 1 good comment), and only affects our test suite so I will merge it now. Plus, it's blocking some other changes I have in the pipeline like https://github.com/gopherjs/vecty/pull/180.

Issue #179, opened 2017-12-04
Consider changing (*HTML).Node return type from *js.Object to interface{}


slimsag:
Today, the [`(*HTML).Node`](https://godoc.org/github.com/gopherjs/vecty#HTML.Node) method returns `*js.Object`. For right now, this works well. But in the future, it could be bad for several reasons:

1. If server-side-rendering (#23) happens eventually, we may want this to return some other data type (not sure what).
2. When [WebAssemby is supported in Go](https://github.com/golang/go/issues/18892), it probably won't expose JS through the existing github.com/gopherjs/gopherjs/js package. Or, at the very least, it probably won't be at that exact location (but rather in the stdlib somewhere). The return type will be different.
3. It's possible that once we figure out testing (#29), we might want it to return some mocked DOM object type.

Now, one could argue that instead of overloading `Node()` to return an `interface{}` and making the caller deal with the type, we could instead use more methods (`ServerNode()`, `WebAssemblyNode()`, `TestNode()`, etc.) but I think this would lead to us pulling in too many unwanted dependencies for specific use cases of Vecty, to the point where changing the return type to `interface{}` is probably warranted.

If nobody objects, we can change the return type to `interface{}` and add an entry in `docs/CHANGELOG.md` to reflect it.

pdf:
My preference would be to leave it as is for the time being - if we have a concrete reason to change it, we'll have more understanding of why.  Making people cast stuff without a good reason is not doing them any favours.

slimsag:
Fair enough, I agree with that @pdf

slimsag:
Something to figure out before 1.0, but after Go gets wasm support and after we figure out what to do about server-side-rendering.

slimsag:
https://github.com/gopherjs/vecty/tree/sg/node-interface for if we decide to do this eventually. (wrote it before your comment)

shurcooL:
This sounds like a change that you should defer doing until the very moment you actually need this. Doing it sooner has no benefits IMO, but has downsides in that you might end up changing your mind or gain additional insight by the time you need it.

PR #180, opened 2017-12-04
 add RenderInto (fixes #81)

### LINKED ISSUE #81: vecty.Render into non-body component
Currently Render is restricted to a body component; i.e. Vecty must own the whole page effectively. We should allow rendering directly into an element that we own, e.g. `vecty.RenderInto("elem-id")` or something.

slimsag:
Based on #178, which should be reviewed first. Only the last commit here is part of this PR itself.

Fixes #81


slimsag:
Given discussion in #179 the API here should probably not accept `interface{}`

slimsag:
Closing since I need to update this anyway.

PR #186, opened 2018-01-07
It is only safe to consider keyed children stable based on node equality

### LINKED ISSUE #185: Something wrong with Keyer interface?
When changing the order of a `vecty.List` the elements don't always get reordered correctly.

To recreate the issue sort the `Category` column twice so that it is in descending order. In the console you'll see that the slice has been sorted correctly, however, the DOM elements haven't been reordered correctly.

```go
package main

import (
	"fmt"
	"sort"

	"github.com/gopherjs/vecty"
)

type root struct {
	vecty.Core
	Results       []result
	sortColumn    string
	sortDirection string
}

type row struct {
	vecty.Core
	Result result
}

type result struct {
	Domain   string
	Category string
}

func main() {
	vecty.SetTitle("Vecty Sort")
	vecty.AddStylesheet("https://cdn.jsdelivr.net/npm/semantic-ui@2/dist/semantic.min.css")
	vecty.RenderBody(&root{
		Results: []result{
			result{Domain: "bing.com", Category: "Search"},
			result{Domain: "facebook.com", Category: "Social Media"},
			result{Domain: "google.com", Category: "Search"},
			result{Domain: "instagram.com", Category: "Social Media"},
			result{Domain: "myspace.com", Category: "Social Media"},
			result{Domain: "somethingawful.com", Category: "Social Media"},
			result{Domain: "twitter.com", Category: "Social Media"},
			result{Domain: "pinterest.com", Category: "Social Media"},
			result{Domain: "reddit.com", Category: "Social Media"},
			result{Domain: "snapchat.com", Category: "Social Media"},
		},
	})
}

func (r *root) Sort(column string) func(*vecty.Event) {
	return func(e *vecty.Event) {
		if r.sortColumn == column {
			if r.sortDirection == "down" {
				r.sortDirection = "up"
			} else {
				r.sortDirection = "down"
			}
		} else {
			r.sortColumn = column
			r.sortDirection = "up"
		}

		println(r.sortColumn, r.sortDirection)

		sort.Slice(r.Results, func(a, b int) bool {
			if r.sortDirection == "down" {
				a, b = b, a
			}
			switch r.sortColumn {
			case "domain":
				return r.Results[a].Domain < r.Results[b].Domain
			case "category":
				return r.Results[a].Category < r.Results[b].Category
			}
			panic("sortColumn not found")
		})

		for _, result := range r.Results {
			fmt.Printf("%#v\n", result)
		}

		vecty.Rerender(r)
	}
}

func (r *root) Render() vecty.ComponentOrHTML {
	rows := vecty.List{}
	for _, result := range r.Results {
		rows = append(rows, &row{
			Result: result,
		})
	}

	return elem.Body(
		elem.Table(
			vecty.Markup(
				vecty.Class("ui", "selectable", "striped", "unstackable", "single", "line", "table"),
				vecty.Style("margin-bottom", "20px"),
			),
			elem.TableHead(
				elem.TableRow(
					elem.TableHeader(
						vecty.Markup(
							event.Click(r.Sort("domain")),
							vecty.Style("cursor", "pointer"),
						),
						vecty.Text("Domain"),
						vecty.If(
							r.sortColumn == "domain",
							elem.Italic(
								vecty.Markup(
									vecty.Class("caret", r.sortDirection, "icon"),
								),
							),
						),
					),
					elem.TableHeader(
						vecty.Markup(
							event.Click(r.Sort("category")),
							vecty.Style("cursor", "pointer"),
						),
						vecty.Text("Category"),
						vecty.If(
							r.sortColumn == "category",
							elem.Italic(
								vecty.Markup(
									vecty.Class("caret", r.sortDirection, "icon"),
								),
							),
						),
					),
				),
			),
			elem.TableBody(
				rows,
			),
		),
	)
}

func (r *row) Key() interface{} {
	return r.Result.Domain
}

func (r *row) Render() vecty.ComponentOrHTML {
	return elem.TableRow(
		elem.TableData(
			vecty.Text(r.Result.Domain),
		),
		elem.TableData(
			vecty.Text(r.Result.Category),
		),
	)
}
```

pdf:
Because we must move children by inserting them before another node (so
children only ever move up, never down), we must only consider them
stable by node position, not by children index.

Fixes #185

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -428,23 +428,8 @@ func (h *HTML) reconcileChildren(prev *HTML) (pendingMounts []Mounter) {
 			prevChild = prev.children[i]
 		}
 		// Find previous keyed sibling if exists, and mutate from there.
-		var stableKey bool
 		if hasKeyedChildren {
 			if prevKeyedChild, ok := prev.keyedChildren[nextKey]; ok {
-				// If the previous node rendered at this position matches the
-				// keyed previous render node, we have a stable key, and will
-				// replace later instead of inserting.
-				var prevPositionRender, prevKeyedRender *HTML
-				if _, isList := prevChild.(KeyedList); !isList {
-					prevPositionRender = extractHTML(prevChild)
-				}
-				if _, isList := prevKeyedChild.(KeyedList); !isList {
-					prevKeyedRender = extractHTML(prevKeyedChild)
-				}
-				if prevPositionRender != nil && prevKeyedRender != nil && prevPositionRender.node == prevKeyedRender.node {
-					stableKey = true
-				}
-
 				prevChild = prevKeyedChild
 				// Delete the matched key from the previous index map so that
 				// we can remove any dangling children.
@@ -494,12 +479,16 @@ func (h *HTML) reconcileChildren(prev *HTML) (pendingMounts []Mounter) {
 			prevChild = nil
 		}
 
-		// If we're keyed and not stable, find the next DOM node from the
-		// previous render to insert before, for reordering.
-		var insertBeforeKeyedNode jsObject
-		if hasKeyedChildren && !stableKey {
+		// If we're keyed, find the next DOM node from the previous render to
+		// insert before, for reordering.
+		var (
+			insertBeforeKeyedNode jsObject
+			stableKey             bool
+		)
+		if hasKeyedChildren {
 			insertBeforeKeyedNode = h.lastRenderedChild.nextSibling()
-			// If the next node is our old node, mark key as stable.
+			// If the next node is our old node, mark key as stable, to avoid
+			// unnecessary insertion.
 			if prevChildRender != nil && prevChildRender.node == insertBeforeKeyedNode {
 				stableKey = true
 				insertBeforeKeyedNode = nil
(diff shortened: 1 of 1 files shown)


pdf:
The removed code makes some assumptions that are only true in some re-ordering scenarios.  It may be possible to improve performance here, but not without adding complexity, and I'm not even sure the performance gains are worth it, except possibly for very large lists.

BTW @slimsag I tried adding tests to cover keyed children, but I ran into some issues with the new test framework, that I'll file in a separate issue.

pdf:
I can add the tests that I do have, but the results not entirely accurate because the `wanted` calls produced from the test suite don't match reality.

slimsag:
Thanks for sending this fix @pdf ! I think that having this code be simpler (like this) for now is good, and if performance is a problem in certain situations we can always revisit optimization later. :)

Apologies for the delayed review on this!

> I can add the tests that I do have, but the results not entirely accurate because the wanted calls produced from the test suite don't match reality.

Yeah, better to not do that for now if the wanted calls don't match reality. I will follow up on #187 soon.

Merging without the tests for now to fix #185 (cc @hanleym) :) thanks again @pdf!

Issue #136, opened 2017-09-02
Switch from MDN-doc-scraper to custom open data format


slimsag:
Right now our generated packages (`elem`, `prop`, etc) are created by scraping the MDN documentation website and pulling relevant information. At first, this gave us good coverage of the entire DOM API and worked well, but now it is a waste of time and gives us very inaccurate results.

## MDN Background

As time went on, I noticed that a significant portion of the pages on the MDN are in very inconsistent formats which makes the docs extremely hard to scrape for information accurately:

- Some pages use a table layout, while others use separate headers to represent each object method/property.
- Many pages have incorrect/inconsistent icon markers for experimental/deprecated/donotuse features.
- Documentation strings do not start in any consistent English way (`"background" -- background color`, `"background" alters the background color`, `"background" property`, etc.) which means our godocs read very strangely.

To resolve the above issues, I spent upwards of 80+ hours contributing to the MDN in order to resolve these issues. I found the best layouts on the most popular MDN pages, and ensured other pages follow that same style consistently in page layout and wording.

Unfortunately, this was mostly in vain. The MDN is a bit like the wild-west: anyone can make changes if they have a GitHub account, without any peer review(!), and anyone can revert changes without any peer review(!).

**Although I made over ~85 pages use a consistent layout to the rest of the MDN and clearly documented my changes as doing this, almost 18 of those pages were reverted by another MDN contributor without any reason mentioned in the history.** I tried to reach this contributor via the IRC channel and mailing lists, as he had no public contact information, but came up with still no way to contact this contributor after several weeks.

With no way to contact this contributor, I made attempts to change the page layout on a few of those pages again and directly mentioned in the changelog that his revert made the page not follow the consistent style used on other popular MDN pages, and that I was trying to adopt a consistent MDN format. Again, the changes were reverted.

## Better approach

The MDN's content license is permissive enough for us to use their documentation in our godocs, and so I think we should use an alternative method of generating our packages from (initially) the MDN documentation.

What this would look like is creating a separate Vecty repository, maybe `github.com/vecty/webdoc` with some type of file format (YAML, XML, etc) that documents individual web APIs (objects, data types, function signatures, docstrings, etc) for use in our generators.

This is similar in concept to how [Glow](https://github.com/go-gl/glow), a Go OpenGL binding generator I have worked on, operates.

pdf:
[...]

PR #195, opened 2018-04-15
Update selectors for querying events from mozilla

gernest:
It seems the page https://developer.mozilla.org/en-US/docs/Web/Events
has changed. This commit Updates the selector to reflect this.


event/generate.go
```diff
@@ -135,7 +135,7 @@ func main() {
```
shurcooL: I suspect it's a good idea to update `nameMap` above so the generated names have proper casing:

https://github.com/gopherjs/vecty/blob/19a11ec3acc3ae6c2e095b1ace2b7134c5f28c10/event/generate.go#L22-L129

gernest: Can you please expand on what you mean by proper casing? The goal of this PR is just to fix the selector to properly query the page. I believe other changes can be more meaningful in a separate PR.

shurcooL: Sure. This PR currently adds some new functions to `event` package, like:

- `Devicechange`
- `Messageerror`
- `Slotchange`
- etc.

According to Go style conventions (https://golang.org/s/style#mixed-caps and https://golang.org/s/style#initialisms), identifiers should use MixedCaps casing. So the new functions should be named:

- `DeviceChange`
- `MessageError`
- `SlotChange`
- etc.

`nameMap` provides a mapping from HTML attribute names to proper Go style names. By adding new entries there, you can ensure the functions that are added as part of this PR follow Go style.

If your goal is not to add any new functions to `event` package as part of this PR, you'll need to make changes. That's not currently the case.

gernest: Thanks for the explanation. I will add the mapping, it will be weird just submitting the changes in the generator file without the updated generated file.

gernest:
@shurcooL I updated the nameMap. Hope I didn't miss anything.

event/generate.go
```diff
@@ -30,6 +30,7 @@ func main() {
 		"animationend":             "AnimationEnd",
 		"animationiteration":       "AnimationIteration",
 		"animationstart":           "AnimationStart",
+		"appinstalled":             "AppInstalled",
```
shurcooL: I'm not sure if "App" should be considered a full word, or if it's short for "Application". If the former, this is good. If the latter, this should be mapped to "ApplicationInstalled", similar to how "dblclick" is mapped to "DoubleClick" below (rather than to "DblClick").

Given that the event is [described](https://developer.mozilla.org/en-US/docs/Web/Events/appinstalled) as (emphasis mine):

> The `appinstalled` event of the Web Manifest API is fired when the browser has successfully installed a page as an **application**.

I'm leaning slightly towards "ApplicationInstalled".

I'll defer to @slimsag on this.

slimsag: I agree, `ApplicationInstalled` is preferred here.

event/event.gen.go
```diff
@@ -376,7 +382,7 @@ func DurationChange(listener func(*vecty.Event)) *vecty.EventListener {
 
 // Emptied is an event fired when the media has become empty; for example, this
 // event is sent if the media has already been loaded (or partially loaded),
-// and the load() method is called to reload it.
+// and the load() method is called to reload it.
```
slimsag: We should figure out what changed here. This may be the MDN using another weird space unicode character, and if so, we should swap it out in the generator with a plain ASCII space character.

slimsag: Probably copying https://github.com/gernest/vecty/blob/5f529e0c4616b201de3fc21c4c8873c92862bf89/elem/generate.go#L158-L160 into this generator would do the trick.

event/event.gen.go
```diff
@@ -53,14 +61,6 @@ func AudioEnd(listener func(*vecty.Event)) *vecty.EventListener {
 	return &vecty.EventListener{Name: "audioend", Listener: listener}
 }
 
-// AudioProcess is an event fired when the input buffer of a
-// ScriptProcessorNode is ready to be processed.
-//
-// https://developer.mozilla.org/docs/Web/Events/audioprocess
-func AudioProcess(listener func(*vecty.Event)) *vecty.EventListener {
```
slimsag: It seems like this has gotten removed by the generator even though it is still a valid event (not deprecated etc). We'll need to figure out / fix why this got removed before we can merge this.

event/event.gen.go
shurcooL: > We should figure out what changed here.

I already checked. MDN changed a non-breaking whitespace (U+00A0) for a normal whitespace (U+0020).

image

Copying `desc = strings.Replace(desc, "\u00a0", " ", -1)` might still be desirable if you really want to eliminate those non-breaking whitespaces, in case they come up in the future or elsewhere.

However, I think having non-breaking whitespace in documentation isn't a big deal. I don't see a reason to actively avoid them.

slimsag: Got it. This change LGTM as-is then.

Unrelated to this change, but I do want to avoid non-breaking whitespace in documentation because it easily leads to changes like this where there is uncertainty about why a line has changed.

event/event.gen.go
gernest: @slimsag apologies for the late feedback, I was  in a hospital , I feel better now. I t seems the cause is the `.icon-thumbs-down-alt` class see https://github.com/gopherjs/vecty/pull/195/files#diff-7f46ba42e4ad2574c9cbe6836be77d61L140  I don't know why the logic is there. Maybe the docs added that icon for some reason.

gernest:
@shurcooL I fixed the nameMap for `appinstalled`. Please let me know if further changes are requires.

event/event.gen.go
slimsag: Sorry to hear you were not feeling well! Hope you're feeling better.

I just had a chance to look at this, and it looks like the problem is just more inconsistency on the MDN -_-

image

The `audioprocess` event is not marked as deprecated, but the data type it events `AudioProcessingEvent` is deprecated. Obviously that is silly because that means both are deprecated, so we're doing the right thing here by removing this (since we don't keep deprecated items in here). Just another reason to solve #136 soon.

PR #197, opened 2018-04-18
Panic usefully when (*HTML).Node() is called too early

### LINKED ISSUE #196: (*HTML).Node() panic if called before Mount phase
I assume the rendered *HTML's JS object simply isn't available until a certain point in a component's lifecycle (Mount), that's fine. The issue is the following panic at https://github.com/gopherjs/vecty/blob/19a11ec3acc3ae6c2e095b1ace2b7134c5f28c10/dom.go#L153
```
Error: interface conversion: interface is nil, not vecty.wrappedObject
```

Returning `nil` instead, if the *js.Object isn't available yet, and documenting (*HTML).Node() with that fact seems appropriate. I'll submit a PR if there's no objections to that.

bzub:
Fixes #196

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -149,8 +149,15 @@ type HTML struct {
 	lastRenderedChild *HTML
 }
 
-// Node returns the underlying JavaScript Element or TextNode.
-func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
+// Node returns the underlying JavaScript Element or TextNode. Node panics
+// when called before the associated component's Mount phase.
+func (h *HTML) Node() *js.Object {
+	if h.node == nil {
+		panic("(*HTML)Node() called before node creation. " +
+			"The component should be mounted before using Node().")
+	}
+	return h.node.(wrappedObject).j
+}
 
 // Key implements the Keyer interface.
 func (h *HTML) Key() interface{} {
--- a/dom.go
+++ b/dom.go
@@ -149,12 +149,13 @@ type HTML struct {
 	lastRenderedChild *HTML
 }
 
-// Node returns the underlying JavaScript Element or TextNode. Node panics
-// when called before the associated component's Mount phase.
+// Node returns the underlying JavaScript Element or TextNode.
+//
+// It panics if it is called before the DOM node has been attached, i.e. before
+// the associated component's Mounter interface would be invoked.
 func (h *HTML) Node() *js.Object {
 	if h.node == nil {
-		panic("(*HTML)Node() called before node creation. " +
-			"The component should be mounted before using Node().")
+		panic("vecty: cannot call (*HTML).Node() before DOM node creation / component mount")
 	}
 	return h.node.(wrappedObject).j
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: go
 go:
-  - 1.9.x
+  - 1.10.x
 install:
   - nvm install node
   - npm install -g source-map-support


pdf:
Can't change Go versions without changing GopherJS versions.

slimsag:
> Can't change Go versions without changing GopherJS versions.

I think he was trying to fix the existing Travis failure actually

dom.go
```diff
@@ -149,8 +149,15 @@ type HTML struct {
...
 }
 
-// Node returns the underlying JavaScript Element or TextNode.
-func (h *HTML) Node() *js.Object { return h.node.(wrappedObject).j }
+// Node returns the underlying JavaScript Element or TextNode. Node panics
+// when called before the associated component's Mount phase.
+func (h *HTML) Node() *js.Object {
+	if h.node == nil {
+		panic("(*HTML)Node() called before node creation. " +
+			"The component should be mounted before using Node().")
```
slimsag: A few things:

1. All Vecty panics should start with `vecty: `, akin to how some Go stdlib packages do it.
2. The string concatenation here is needless.
3. `(*HTML)Node()` is more correctly written as `(*HTML).Node()` I think.
4. It's weird to reference it as `(*HTML)Node()` at the start of the panic message, but as just `Node()` later on in the panic message.

I suggest writing this as just:

```Go
panic("vecty: cannot call (*HTML).Node() before node creation / component mount")
```

bzub:
Thanks for the comments. It's good to learn how to keep things consistent too.

If it looks good now I will squash commits. I was wondering if the travis change should be a separate commit though?

slimsag:
I restarted the travis build since it was still failing, but I don't understand why we are getting those permission errors opening testdata files?

slimsag:
> If it looks good now I will squash commits. I was wondering if the travis change should be a separate commit though?

In general, yeah, a separate commit for travis changes is preferred in my experience. In our case here (for this project, Vecty) we do squash merges of all GitHub PRs so you don't need to do any squashing at all. Your entire PR will end up being merged as a single commit regardless.

slimsag:
The travis failure:

```
--- FAIL: TestHTML_reconcile_std (0.12s)
    --- FAIL: TestHTML_reconcile_std/text_identical (0.01s)
    	test.918843179:35080: open testdata/TestHTML_reconcile_std__text_identical.want.txt: permission denied
    --- FAIL: TestHTML_reconcile_std/text_diff (0.00s)
    	test.918843179:35080: open testdata/TestHTML_reconcile_std__text_diff.want.txt: permission denied
    --- FAIL: TestHTML_reconcile_std/properties (0.02s)
        --- FAIL: TestHTML_reconcile_std/properties/diff (0.00s)
        	test.918843179:35080: open testdata/TestHTML_reconcile_std__properties__diff.want.txt: permission denied
        --- FAIL: TestHTML_reconcile_std/properties/remove (0.01s)
        	test.918843179:35080: open testdata/TestHTML_reconcile_std__properties__remove.want.txt: permission denied
        --- FAIL: TestHTML_reconcile_std/properties/replaced_elem_diff (0.00s)
        	test.918843179:35080: open testdata/TestHTML_reconcile_std__properties__replaced_elem_diff.want.txt: permission denied
        --- FAIL: TestHTML_reconcile_std/properties/replaced_elem_shared (0.00s)
        	test.918843179:35080: open testdata/TestHTML_reconcile_std__properties__replaced_elem_shared.want.txt: permission denied
```

No clue what has changed here or why this would be failing. It used to work.

I'm happy to merge this regardless of the above, though, I don't think it's caused by your PR.

PR #200, opened 2018-04-30
Always call Render() in the persistent component

damz:
(My understanding of the code base is limited, my apologies in advance if this is way off-base.)

I understand that Vecty aims at supporting a concept of "persistent" component, in which the initial instance of a component is re-used at each rendering... and it does seem like a desirable property for non-trivial applications.

Unfortunately, it seems to fail pretty quickly. Consider [this simple example](https://gist.github.com/damz/5609c4a65535627cf7f71adc2c797d5e):

* The rendering is currently done in the _new_ component, so the event handlers are bound to the new component, and mutate the new component
* But when the body is re-rendered, it reuses the _persistent_ component, overriding the state that was mutated by the event handler

This MR changes the render logic to always call the `Render()` method in the _persistent_ component.


bzub:
Quick note on the CI failure, it's a GopherJS issue that will hopefully be fixed soon in https://github.com/gopherjs/gopherjs/pull/810

slimsag:
@damz I won't have time to review this properly until probably this weekend or next.

Can you add a test that fails before your change and passes afterwards, though? We will need that to merge this type of change in (and it will help a lot with understanding the problem clearly)

pdf:
I'm not sure about the proposed change, but there is a bug here - persistence does appear to be broken currently if a component is rendered inside a HTML element, but works if the component is returned directly from a `Render`.  I suspect this means a change should occur inside `reconcile`, rather than `renderComponent`, but I also won't have a chance to review this week.

To illustrate, this works:
https://gist.github.com/pdf/b1d8109edf90d4b3a7f049c40a6f4a59/5adf4f7dde34009003c6ec003ebc9791b65867c1
but this doesn't:
https://gist.github.com/pdf/b1d8109edf90d4b3a7f049c40a6f4a59/5dec93e622465468634b487b9e88eeb28ec9a1f8

damz:
@slimsag I added a test that fails on the current master, but passes with the fix I suggesting here.

pdf:
Thanks for the response @damz, though I'm not entirely sure I follow your reasoning.  For reference:
- `renderComponent` is called recursively on components, this is expected.
- `copyProps` only copies fields from the `next` instance that are tagged with `vecty:"prop"`, this is to allow setting fields from the parent's render context, when persistence is in effect.
- `copyComponent` is called to populate `Context().prevRenderComponent`, which is only used to allow components implementing the optional `RenderSkipper` interface to compare component field values from the previous render, to determine whether to skip this render - this is required especially to allow persistence to be usable in combination with `RenderSkipper`.

In any case, I think the below might be the cleanest fix for this issue, though I'll need some time to confirm. I feel like we were actually doing something very much like this at some point, so I'll have to dig through the history when I have the head-space, to work out what exactly happened here.

```diff
diff --git a/dom.go b/dom.go
index acce31a..889e9ae 100644
--- a/dom.go
+++ b/dom.go
@@ -1012,6 +1012,8 @@ func renderComponent(next Component, prev ComponentOrHTML) (nextHTML *HTML, skip
                // what properties the parent has specified during SkipRender/Render
                // below.
                copyProps(next, prevComponent)
+               // Persist the previous component across renders
+               next = prevComponent
        }
 
        // Before rendering, consult the Component's SkipRender method to see if we
```

damz:
@pdf Your approach seems to work too, so I adopted it here.

The only downside of it that I can see is that we call `copyComponent()` "on each render", instead of "only once in the initial render" in my initial patch.

pdf:
> copyProps(next, prevComponent) actually does two things

Ah, right you are, shows how good my memory is.  I also recall now why we did it that way - the persisted component needs to propagate out through the whole render chain. That being the case, I'll need more time to actually dig into this, as that should cause `Render` to be called on the correct instance, since `next` should effectively equal `prev` after a call to `copyProps`.  Looking at `copyProps`, I actually wonder if there may be an upstream change involved here - will have to test against older go/gopherjs versions when I get a chance, to confirm.

> we call copyComponent() "on each render", instead of "only once in the initial render"

This is absolutely necessary - `RenderSkipper` is designed to allow a user to skip a render if they can determine that a component has not changed between renders in a way that should trigger a new render (so they can optimise perf).  To do so ,they need access to the component as it existed on the previous render - because components are persisted, without the copy they would only see the current version, and could not perform any comparisons on field values.

damz:
> That being the case, I'll need more time to actually dig into this, as that should cause `Render` to be called on the correct instance, since `next` should effectively equal `prev` after a call to `copyProps`.

That's the problem right there. Being equal is not the same thing as being the same object. Yes, after `copyProps` the two objects are equal, but they are not the same object, so event handlers end up modifying the wrong instance.

> To do so ,they need access to the component as it existed on the previous render - because components are persisted, without the copy they would only see the current version, and could not perform any comparisons on field values.

That property is not changed by my patch, and it is actually tested in the test suite. The `next` version and the `prev` version are still two different objects.

What's happening anyway is that `NextRender` is called on `next` (which is typically a brand new object created by `Render()` of the parent). And it gets passed `prev` which has to have the same values as the component that was previously rendered.

pdf:
> Being equal is not the same thing as being the same object.

We're essentially doing this: `*next = *prev`, but via reflect, since we can't know the concrete types up front do do the indirect.

> That property is not changed by my patch, and it is actually tested in the test suite.

I think the tests only cover a single `Rerender` iteration, which is why your proposed change did not fail them (that change would result in the original state being passed to `SkipRender` every render, rather than the previous state).  We should probably modify the tests to perform two `Rerender`s, to ensure that `Context().prevRenderComponent` is successfully updated every render.

damz:
> We're essentially doing this: `*next = *prev`, but via reflect, since we can't know the concrete types up front do do the indirect.

Absolutely. And this is a _shallow_ copy.

> I think the tests only cover a single Rerender iteration, which is why your proposed change did not fail them (that change would result in the original state being passed to SkipRender every render, rather than the previous state).

I see. You are right, what we need here is to compare "current (persistent) state" vs "state at the time of the last render".

Anyway, I'm not disagreeing. I cannot see anything wrong with the current patch.

pdf:
Okay, I believe I understand how we got here - we now explicitly store the previous component on the parent's `children` slice during the current `reconcileChildren`, and the associated changes broke persistence.  In the past, the clone in `copyProps` propagated the previous component's data up to the parent, but that's now useless, as the next value is thrown away.

Now that I understand it, I'm pretty confident that current change is correct, and we can also drop the clone in `copyProps`, to save ourselves a copy there:

```diff
diff --git a/dom.go b/dom.go
index acce31a..bb1f86b 100644
--- a/dom.go
+++ b/dom.go
@@ -940,7 +940,7 @@ func copyComponent(c Component) Component {
 }
 
 // copyProps copies all struct fields from src to dst that are tagged with
-// `vecty:"prop"`, then sets *src = *dst.
+// `vecty:"prop"`
 //
 // If src and dst are different types or non-pointers, copyProps panics.
 func copyProps(src, dst Component) {
@@ -965,11 +965,6 @@ func copyProps(src, dst Component) {
                        df.Set(sf)
                }
        }
-       // Effectively set *src = *dst. This allows the replacement to propagate out of
-       // the calling function, which is needed for components which themselves return
-       // components (see how renderComponent works for components that return
-       // components).
-       s.Elem().Set(d.Elem())
 }
 
 // render handles rendering the next child into HTML. If skip is returned,
```

damz:
> Okay, I believe I understand how we got here - we now explicitly store the previous component on the parent's children slice during the current `reconcileChildren`, and the associated changes broke persistence.

That also means that the issue with "`Render()` returns a `Component` directly" is because we are not doing the same thing in `renderComponent`:

```
@@ -1043,6 +1043,9 @@ func renderComponent(next Component, prev ComponentOrHTML) (nextHTML *HTML, skip
                if skip {
                        return nextHTML, skip, pendingMounts
                }
+               if prevComponent, ok := prevRender.(Component); ok && sameType(v, prevComponent) {
+                       nextRender = prevRender
+               }
        case *HTML:
                if v == nil {
                        // nil renders are translated into noscript tags.
```

damz:
This PR now has two different, but related fixes:

* Always call `Render()` on the persistent component;
* Persist components when they are returned directly by `Render()`

It also:

* Fix the CI that is broken because it doesn't specify a major version for NodeJS;
* Adds two tests that both fail before these fixes;
* Removes the now unecessary `*src = *dst` hack in `copyProps()`

How does that look?

pdf:
LGTM

damz:
@slimsag Any feedback on this?

damz:
Gentle bump.

slimsag:
Apologies for the delay. I haven't had time to review this fully, but took a quick glance just now and given the added tests + pdf's approval, I'll merge now. Thanks a ton for sending this! :)

PR #201, opened 2018-05-01
Document prevRenderComponent and prevRender

damz:
Documentation follow up to #200, clarifying what `Core.prevRenderComponent` and `Core.prevRender` actually are.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -12,8 +12,8 @@ var batch = &batchRenderer{idx: make(map[Component]int)}
 // Core implements the Context method of the Component interface, and is the
 // core/central struct which all Component implementations should embed.
 type Core struct {
-	prevRenderComponent Component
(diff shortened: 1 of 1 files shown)


dom.go
```diff
@@ -12,8 +12,8 @@ var batch = &batchRenderer{idx: make(map[Component]int)}
 // Core implements the Context method of the Component interface, and is the
 // core/central struct which all Component implementations should embed.
 type Core struct {
-	prevRenderComponent Component
-	prevRender          ComponentOrHTML
+	prevRenderComponent Component       // A shallow copy of the component at the time of the last render.
```
slimsag:
```suggestion
	prevRenderComponent Component       // A copy of the component at the time of the last render.
```

slimsag:
Stale PR; closing.

Issue #194, opened 2018-04-08
All DOM properties must be comparable


bzub:
When using `vecty.Property()`, a panic is caused when the value is an uncomparable type.

At https://github.com/gopherjs/vecty/blob/19a11ec3acc3ae6c2e095b1ace2b7134c5f28c10/dom.go#L253
```
comparing uncomparable type func() error
```

One solution could be to provide an interface the user can satisfy to prove equality of a property.

pdf:
> One solution could be to provide an interface the user can satisfy to prove equality of a property.

I'm not sure how that would work, but we could probably check the type and unconditionally apply uncomparable types.

slimsag:
Could someone clarify for me which properties are allowed to be non-comparable types?

My initial assumption was that only valid DOM properties could be specified with `vecty.Property` but I see the referenced commit is using properties like `vecty-material-ripple` and I'm curious what the use case is. I guess the use case is just associating arbitrary Go data with elements?

bzub:
> I guess the use case is just associating arbitrary Go data with elements?

That's correct in my case. More specifically it's for including Go data from outside of a component that needs to store the component's rendered HTML at the Apply phase, and then use the *js.Object provided by Node() at the Mount phase of the component's lifecycle. I wanted to do this truely arbitrarily, so this type of Applyer could be used with any component (not there yet in the commit/code mentioned earlier). The *js.Object of the element is needed because I'm using a binding to an external JS library -- besides that requirement I'm very open to alternative approaches. Using an Applyer seems very intuitive so far.

pdf:
> Could someone clarify for me which properties are allowed to be non-comparable types?

Whilst user-defined properties are somewhat frowned upon they are valid, and may contain arbitrary data. Though I can't think of any standardised properties that accept incomparable types off the top of my head.

I can't say I'm much of a fan of using the DOM as storage like this - I'd much rather see this sort of thing stored in the virtual DOM as a field on a `vecty.Component`.

damz:
@bzub I think what you really need here is an applier that can subscribe to mount/unmount events of an `*vecty.HTML`. See #202.

bzub:
That's exactly what I've shifted toward @damz. So I think this issue is primarily about adding documentation and/or a more useful error message so users doing weird things with vecty.Property() can save some time when they encounter the uncomparable type situation.

PR #202, opened 2018-05-21
Allow `vecty.HTML` to participate in the lifecycle of the node (for discussion)

damz:
There are very few differences between `*vecty.HTML` and `vecty.Component` right now. One of them is that a component has a `Render` method, while a `*vecty.HTML` is already fully rendered. The other is that `*vecty.HTML` cannot participate in its own lifecycle: it doesn't have a way to react to being mounted or unmounted.

Changing this is actually almost trivial, because we _already_ check if any element of the hierarchy is a `vecty.Mounter` or a `vecty.Unmounter`. We just need to implement `HTML.Mount` and `HTML.Unmount`, and wire them into an applier.

### DIFF
--- a/dom.go
+++ b/dom.go
@@ -147,6 +147,9 @@ type HTML struct {
 	// lastRendered child tracks the last child that was rendered, across List
 	// boundaries.
 	lastRenderedChild *HTML
+
+	mounters   []func(h *HTML)
+	unmounters []func(h *HTML)
 }
 
 // Node returns the underlying JavaScript Element or TextNode.
@@ -158,6 +161,26 @@ func (h *HTML) Node() *js.Object {
 		panic("vecty: cannot call (*HTML).Node() before DOM node creation / component mount")
 	}
 	return h.node.(wrappedObject).j
+}
+
+func (h *HTML) Mount() {
+	if h.mounters == nil {
+		return
+	}
+
+	for _, m := range h.mounters {
+		m(h)
+	}
+}
+
+func (h *HTML) Unmount() {
+	if h.unmounters == nil {
+		return
+	}
+
+	for _, u := range h.unmounters {
+		u(h)
+	}
 }
 
 // Key implements the Keyer interface.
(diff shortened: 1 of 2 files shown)


pdf:
The reconciliation code is missing here, for mutating one element to another when elements are replaced or moved in the DOM.

However, I'd like to understand why using a component is not acceptable, can you shed some light?

damz:
> However, I'd like to understand why using a component is not acceptable, can you shed some light?

A `vecty.Component` right now does not have any access to the underlying DOM node. (That's something that could be fixed separately.)

But more than that, reducing the (somewhat arbitrary) differences between `vecty.HTML` and `vecty.Component` makes everything more flexible. It allows to easily compose behaviors, like in @bzub example in #194:

```go
func (c *MyComponent) Render() vecty.ComponentOrHTML {
	return elem.Button(
		vecty.Markup(
			ripple.Ripple{},
			event.Click(func(ev *vecty.Event) {
				// Do something.
			}),
		),
		vecty.Text("Hello, ripple."),
	)
}
```

Of course, you could implement that as a special "button with ripple effect" component, but then this behavior cannot be composed with other behaviors (for example

damz:
> The reconciliation code is missing here, for mutating one element to another when elements are replaced or moved in the DOM.

Actually that part works right now. But it's because _any_ rerender seems to trigger mount/umount on the `vecty.HTML` elements.

pdf:
Okay, I'll need to find some time to think about this some more.

damz:
> The reconciliation code is missing here, for mutating one element to another when elements are replaced or moved in the DOM.

What would be a good criteria here? From the logic of the render tree, it would make sense to me that we would consider two elements as being "the same" element between two renders (i.e. not requiring calling mount/unmount) if:

* Either they are unkeyed elements and they have the same place in the child list (i.e. the first child is "the same" as the first child, etc.)
* Or they are keyed elements and they have the same key

Thoughts?

These two cases degenerate to a single check after render: "the old DOM node is the same as the new DOM node", because keyed elements are reusing their DOM node if it exists. Unfortunately, there is a small bug in there which I fixed in #203

slimsag:
Sorry for the (very) delayed response to this PR. I'm trying to understand the motivation of it, but am having trouble doing so.

> A vecty.Component right now does not have any access to the underlying DOM node. (That's something that could be fixed separately.)

Do you mean something like getting a reference to the DOM node that a component renders? This is in fact possible today, with something like this: https://play.jsgo.io/0ac43e222ce8c53548d8ce1c227a271a9691adee

Maybe the use case you have in mind is more complex than that? (could you provide example code if so?)

> But more than that, reducing the (somewhat arbitrary) differences between vecty.HTML and vecty.Component makes everything more flexible. It allows to easily compose behaviors, like in @bzub example in #194 [...]

I've looked over both #194 and the example code you posted below your message, but I am not able to understand how this allows more easily composing behaviors. What is `ripple.Ripple{}`?

---

Hope you don't mind me asking these questions -- I just want to make absolutely sure that we have strong support for valid use-cases (which means I need to fully understand them :P )

slimsag:
Stale PR; closing for now.

PR #203, opened 2018-05-22
Don't leave dangling DOM nodes when the type of a keyed element changed

damz:
When the type of a keyed element changed:

* `HTML.reconcile` creates a new DOM node
* but `reconcileChildren` still expects that `insertBefore` of the new DOM node will remove the previous DOM node from the tree

This results into dangling nodes still being in the tree after render.


pdf:
Sounds plausible, I'll try and find some time to review.

damz:
@pdf There is a test in there. The difference between before and after the fix is literally just:

```diff
@@ -39,6 +39,4 @@ global.Get("document").Call("createElement", "tag2").Get("classList")
 global.Get("document").Call("createElement", "tag2").Get("dataset")
 global.Get("document").Call("createElement", "tag2").Get("style")
 global.Get("document").Call("createElement", "body").Call("appendChild", jsObject(global.Get("document").Call("createElement", "tag2")))
+global.Get("document").Call("createElement", "tag1").Get("parentNode")
+global.Get("document").Call("createElement", "tag1").Get("parentNode").Call("removeChild", jsObject(global.Get("document").Call("createElement", "tag1")))
 global.Call("requestAnimationFrame", func(float64))
```

damz:
@pdf This is now green if you still want to review.

testsuite_test.go
```diff
@@ -34,8 +34,16 @@ type componentFunc struct {
...
+func (c *componentFunc) Render() ComponentOrHTML {
+	return c.render()
+}
+
+func (c *componentFunc) SkipRender(prev Component) bool {
+	if c.skipRender != nil {
+		return c.skipRender(prev)
+	}
+	return false
+}
```
slimsag: @damz could you undo this change here and instead pass a `skipRender` function in the actual test itself?

I think having this logic be implicit can be misleading / cause us to not notice bugs. Explicitness == better in our case here.

slimsag:
@damz Apologies for the (very) delayed review. This change LGTM. @pdf do you still want to review this before I merge it?

pdf:
I'm afraid I've just had no time to look at it for any edge-cases.  If you're happy though, I'm happy @slimsag

slimsag:
I've applied my inline suggestion in https://github.com/gopherjs/vecty/pull/231 and will merge that (you will still be credited in the authorship info, and feel free to send a PR to add yourself to the CONTRIBUTORS if you like).

PR #205, opened 2018-05-24
elem: Fix lint and generate issues

bzub:
CI currently fails on the `elem` package due to some function comments not being manipulated as needed to satisfy `go vet`. Also, running `go generate` on the elem package generated invalid code because of changes to the MDN website. This PR fixes those issues and touches up some MDN description irregularities.


elem/elem.gen.go
```diff
@@ -788,55 +759,56 @@ func Span(markup ...vecty.MarkupOrChild) *vecty.HTML {
...
+// solely typographical reasons.
 //
 // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sub
 func Subscript(markup ...vecty.MarkupOrChild) *vecty.HTML {
 	return vecty.Tag("sub", markup...)
 }
 
-// Summary is used as a summary, caption, or legend for the content of a
-// <details> element.
+// Summary element specifies a summary, caption, or legend for a <details>
```
slimsag: We'll need to fix this to remove the word `element` or make it a proper sentence somehow.

slimsag:
aside from inline comments, I am fine with merging this.

bzub:
Good catch, there were several of the same type of pattern that weren't matched/corrected. The diff on the last commit seems to only get rid of the extra "element" you commented about. Yay.

PR #214, opened 2018-09-13
Introduce Go Modules

marwan-at-work:


slimsag:
Looks like this is causing some failure on CI

marwan-at-work:
@slimsag not sure what megacheck does but this is what's being returned: 

```
1.68s$ megacheck -tags=js ./...
/home/travis/.gimme/versions/go1.11.linux.amd64/src/runtime/internal/sys/zgoos_linux.go:8:7: GOOS redeclared in this block
couldn't load packages due to errors: runtime, runtime/internal/sys
```

Does it make any sense to you?

slimsag:
megacheck is a very nice linter 😃 

My guess is that megacheck is not playing well with go modules right now. It looks like there is an open issue and a separate `next` branch there for adding go modules support https://github.com/dominikh/go-tools/issues/328

Probably we can try just checking out that `next` branch and seeing if that fixes the problem, otherwise we will have to temporarily disable megacheck which would be sad :/

slimsag:
Fixed conflicts + remaining issues and opened new PR https://github.com/gopherjs/vecty/pull/229

(I retained your authorship information so you still get credit, of course 😄)

PR #231, opened 2019-01-20
Don't leave dangling DOM nodes when the type of a keyed element changed

slimsag:
Supersedes https://github.com/gopherjs/vecty/pull/203 by performing the outstanding work there (retains authorship information of that PR).


dom.go
```diff
@@ -542,6 +539,14 @@ func (h *HTML) reconcileChildren(prev *HTML) (pendingMounts []Mounter) {
 			if m := mountUnmount(nextChild, prevChild); m != nil {
 				pendingMounts = append(pendingMounts, m)
 			}
+
+			if hasKeyedChildren && (prevChildRender != nil && prevChildRender.node == nextChildRender.node) {
```
pdf: This is already inside a `if hasKeyedChildren` block

dom.go
```diff
@@ -439,9 +439,6 @@ func (h *HTML) reconcileChildren(prev *HTML) (pendingMounts []Mounter) {
 		if hasKeyedChildren {
```
pdf: Here's the containing if statement

PR #232, opened 2019-01-21
Add WebAssembly support

### LINKED ISSUE #211: WebAssembly support
This is probably very easy/straightforward to add. We can keep GopherJS support using build tags.

slimsag:
Prior to this PR, Vecty was only usable with GopherJS which would compile Go to JS code.

After this PR, Vecty can be used with:

- GopherJS (browser, testing)
- Go + WebAssembly  (browser, testing)
- Go + Native (for testing Vecty itself, autocomplete in editors etc)

All examples in the `example` directory work under both GopherJS and WebAssembly, and a README is added to describe how to compile for each.

### TinyGo future support

In the future, Vecty will automatically gain support for TinyGo once it supports the subset of reflection and `syscall/js` that we make use of.

See https://github.com/aykevl/tinygo/issues/93 for more discussion around this.

## Prior work

Supersedes the prior PR for this by using build tags instead of exposing a `gopherwasm` API directly. See https://github.com/gopherjs/vecty/pull/215#issuecomment-455829781 for discussion around this.

Fixes #211


slimsag:
Looks like the ecosystem is maturing a bit so we're pretty closer to being able to merge this. Notable things that have happened recently:

- Go 1.12 has been released, so we don't need to advise installing Go tip anymore.
- TinyGo now has support for all reflection APIs we need.
- GopherJS has added `syscall/js` support (thanks to @hajimehoshi) so we no longer need to use gopherwasm in examples and can unify our DOM implementations between GopherJS/WASM.

Remaining work:

- [ ] Test / confirm TinyGo support works
- [ ] Make use of new GopherJS `syscall/js` support / unify DOM implementations.
- [ ] `wasmserve` has a bug only affecting `example/markdown` where it returns:
```
(index):4 GET http://localhost:8080/main.wasm net::ERR_CONTENT_LENGTH_MISMATCH 200 (OK)
```

No other errors, logs, etc. Unsure why. Doesn't affect other examples. cc @hajimehoshi if you have any ideas, otherwise I'll find more minimal case and file/fix upstream.

(Just checkout this branch and `cd example/markdown && wasmserve` if you are keen to reproduce)

aykevl:
I don't think the reflect support in TinyGo is good enough: it still doesn't support inspecting structs for example. However, I have been working on a better map implementation which is a prerequisite to `js.FuncOf` support so there is still progress.

hajimehoshi:
> wasmserve has a bug only affecting example/markdown where it returns:

That's a strange error message I've never seen. I'll take a look later, but now I'm on a business trip, so I cannot guarantee when :-)

slimsag:
Confirmed the wasmserve issue is actually a bug in WSL not wasmserve: https://github.com/hajimehoshi/wasmserve/issues/5

johanbrandhorst:
What is the status of this? I think vecty could move to using `syscall/js` for both WebAssembly and GopherJS now that GopherJS has `syscall/js` too. What do you think?

.travis.yml
```diff
@@ -21,22 +21,23 @@ before_script:
...
+  #- golangci-lint run --exclude 'exported .* should have comment .*or be unexported' ./prop/... ./style/... # https://github.com/gopherjs/vecty/issues/227
+  #- GOOS=js GOARCH=wasm golangci-lint run --build-tags 'js wasm' . ./elem/... ./event/... ./example/...
+  #- GOOS=js GOARCH=wasm golangci-lint run --build-tags 'js wasm' --exclude 'exported .* should have comment .*or be unexported' ./prop/... ./style/... # https://github.com/gopherjs/vecty/issues/227
 
-  # Test with Go compiler and GopherJS compiler.
-  - go test -v -race ./...
-  - gopherjs test -v ./...
+  # Test with Go compiler (under amd64 and wasm architectures) and GopherJS compiler.
+  #- go test -race ./...
+  - GOOS=js GOARCH=wasm go test -exec="$(go env GOROOT)/misc/wasm/go_js_wasm_exec" ./... # https://github.com/golang/go/wiki/WebAssembly#executing-webassembly-with-nodejs
```
johanbrandhorst: FYI there is https://github.com/agnivade/wasmbrowsertest which allows you to test against a browser.

example/README.md
```diff
@@ -0,0 +1,52 @@
+# Building examples
+
+Vecty fully supports all three major Go <-> Web compilers thanks to its minimal dependencies:
+
+- Go 1.12+ WebAssembly support
+- [GopherJS](https://github.com/gopherjs/gopherjs) (Go to JavaScript transpiler)
+- [TinyGo](https://github.com/tinygo-org/tinygo) WebAssembly support (produces tiny binaries, but with tradeoffs)
```
johanbrandhorst: :grimacing: is this right? TinyGo doesn't support all of `reflect` yet. Have you tested it with TinyGo?

example/go.mod
```diff
@@ -1,12 +1,17 @@
 module github.com/gopherjs/vecty/example
 
 require (
-	github.com/gopherjs/gopherjs v0.0.0-20181103185306-d547d1d9531e
-	github.com/gopherjs/vecty v0.0.0-20190121070609-2a1e317a95ae
+	github.com/gopherjs/gopherjs v0.0.0-20190430165422-3e4dfb77656c
+	github.com/gopherjs/gopherwasm v1.1.0
```
johanbrandhorst: Should we be able to get rid of this?

example/todomvc/example.go
```diff
@@ -3,21 +3,21 @@ package main
...
 	"encoding/json"
 
-	"github.com/gopherjs/gopherjs/js"
 	"github.com/gopherjs/vecty"
 	"github.com/gopherjs/vecty/example/todomvc/actions"
 	"github.com/gopherjs/vecty/example/todomvc/components"
 	"github.com/gopherjs/vecty/example/todomvc/dispatcher"
 	"github.com/gopherjs/vecty/example/todomvc/store"
 	"github.com/gopherjs/vecty/example/todomvc/store/model"
+	"syscall/js"
```
johanbrandhorst: Move to stdlib imports?

example/README.md
slimsag: Yeah, I haven't updated this file yet & TinyGo most likely doesn't work. Will fix.

example/go.mod
slimsag: Yep, done

example/todomvc/example.go
slimsag: Good catch, thanks.

.travis.yml
slimsag: Neat, I can see why that'd be useful -- though one of our goals with Vecty is to make `go test` alone (even without GOOS=js) work (initially just for testing Vecty itself, secondarily/not in this PR in projects using Vecty)

slimsag:
Good news! It's finally time for this to be merged. 😃 🎉 

- Consult the updated PR description for details on supported build targets (WebAssembly, GopherJS, native, TinyGo, etc).
- Consult the CHANGELOG for what this means for existing users.
- Consult the new example/README.md for details on building examples for WebAssembly.
- Consult the new README for details on bundle sizes, etc. (hello world is surprisingly small!)

PR #233, opened 2019-01-25
document func event.Timeout

### LINKED ISSUE #226: Add documentation for event.Timeout
```
event/event.gen.go:1138:1: comment on exported function Timeout should be of the form "Timeout ..." (golint)
// (no documentation)
^
```

ilmanzo:
fix #226

### DIFF
--- a/event/event.gen.go
+++ b/event/event.gen.go
@@ -1135,7 +1135,7 @@ func TimeUpdate(listener func(*vecty.Event)) *vecty.EventListener {
 	return &vecty.EventListener{Name: "timeupdate", Listener: listener}
 }
 
-// (no documentation)
+// Timeout event is fired when Progression is terminated due to preset time expiring.
 //
 // https://developer.mozilla.org/docs/Web/Events/timeout
 func Timeout(listener func(*vecty.Event)) *vecty.EventListener {


pdf:
This file is generated, so I think this would need to be overridden somehow in the generator.

ilmanzo:
I see... But running the generator, I get a file without any func definition. Maybe reference page has changed, or parsing is broken ?

```
//go:generate go run generate.go

// Package event defines markup to bind DOM events.
//
// Generated from "Event reference" by Mozilla Contributors,
// https://developer.mozilla.org/en-US/docs/Web/Events, licensed under
// CC-BY-SA 2.5.
package event

import "github.com/gopherjs/vecty"

```

slimsag:
Yeah, the generator is broken at this point due to MDN changes, that doesn't surprise me.

I think making this a hand-written file for now, until we have a better generator source (i.e. https://github.com/gopherjs/vecty/issues/136) is fine. That way we can continue to improve the package and once we do have a better generator, we can use the existing packages as test cases effectively.

slimsag:
@ilmanzo I'd like to merge this as-is, not sure why you closed the PR? Would you be willing to re-send it?

ilmanzo:
Sorry, I understood you were fixing the generator... NP I'll reopen it

Issue #244, opened 2019-07-12
panic: syscall/js: Value.Call: property delete is not a function, got undefined


electricface:
```
panic: syscall/js: Value.Call: property delete is not a function, got undefined wasm_exec.js:41:14
wasm_exec.js:41:14
goroutine 2 [running]: wasm_exec.js:41:14
syscall/js.Value.Call(0x7ff8000000000801, 0x50d1f, 0x6, 0x46c748, 0x1, 0x1, 0x54f320) wasm_exec.js:41:14
	/home/tp/applications/go/src/syscall/js/js.go:321 +0x86 wasm_exec.js:41:14
github.com/gopherjs/vecty.wrappedObject.Delete(...) wasm_exec.js:41:14
	/home/tp/projects/gitlab/deepin-code-release-platform/frontend/src/github.com/gopherjs/vecty/dom_wasmjs_gopherjs.go:97 wasm_exec.js:41:14
github.com/gopherjs/vecty.(*HTML).removeProperties(0x5b9c00, 0x55a460) wasm_exec.js:41:14
	/home/tp/projects/gitlab/deepin-code-release-platform/frontend/src/github.com/gopherjs/vecty/dom.go:305 +0xf wasm_exec.js:41:14
github.com/gopherjs/vecty.(*HTML).reconcileProperties(0x5b9c00, 0x55a460) wasm_exec.js:41:14
	/home/tp/projects/gitlab/deepin-code-release-platform/frontend/src/github.com/gopherjs/vecty/dom.go:218 +0xbe wasm_exec.js:41:14
github.com/gopherjs/vecty.(*HTML).reconcile(0x5b9c00, 0x55a460, 0x55a460, 0x181c000f, 0x1048000b) wasm_exec.js:41:14
	/home/tp/projects/gitlab/deepin-code-release-platform/frontend/src/github.com/gopherjs/vecty/dom.go:207 +0x3a wasm_exec.js:41:14
github.com/gopherjs/vecty.render(0x83380, 0x5b9c00, 0x83380, 0x55a460, 0x5b9c00, 0x85201, 0x5b6db8, 0x1, 0x0) wasm_exec.js:41:14
	/home/tp/projects/gitlab/deepin-code-release-platform/frontend/src/github.com/gopherjs/vecty/dom.go:981 +0x15 wasm_exec.js:41:14
[...]

PR #245, opened 2019-07-15
fix wrappedObject.Delete panic

electricface:
https://github.com/gopherjs/vecty/issues/244

### DIFF
--- a/dom_wasmjs_gopherjs.go
+++ b/dom_wasmjs_gopherjs.go
@@ -26,6 +26,10 @@ var (
 	undefined = wrappedObject{js.Undefined()}
 )
 
+func init() {
+	js.Global().Call("eval", "$vectyDelete = function(obj, key) { delete(obj[key])}")
+}
+
 func funcOf(fn func(this jsObject, args []jsObject) interface{}) jsFunc {
 	return jsFuncImpl{
 		f: js.FuncOf(func(this js.Value, args []js.Value) interface{} {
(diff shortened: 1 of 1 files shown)


flyn-org:
I ran into the same problem, although in my case the symptom ("property delete is not a function, ...") was present consistently on Epiphany (GNOME Web), Firefox, and Chromium. The pull request here fixed the problem for me.

dom_wasmjs_gopherjs.go
```diff
@@ -26,6 +26,10 @@ var (
 	undefined = wrappedObject{js.Undefined()}
 )
 
+func init() {
+	js.Global().Call("eval", "$vectyDelete = function(obj, key) { delete(obj[key])}")
```
slimsag: Sorry for the delayed review. I did review this a while back but waited to see what the resolution was for https://github.com/golang/go/issues/33079 and failed to post that over here.

Go 1.14 has added an explicit `Delete` operator that does not require using `eval` like in this PR: (see https://github.com/golang/go/issues/33079) and using `eval` has negative consequences, like needing a certain CSP (see https://github.com/golang/go/issues/26748)

I think the best option here is to use the new `Delete` method and require Go tip until Go 1.14 is released: https://tip.golang.org/pkg/syscall/js/#Value.Delete

If anyone can update this PR to do that, I can merge this.

slimsag: Sent https://github.com/gopherjs/vecty/pull/256 to fix this.

slimsag:
https://github.com/gopherjs/vecty/pull/256 supersedes this PR, thank you for taking the time to make a contribution though!

PR #249, opened 2019-12-01
Add support for rendering into non-body DOM nodes

### LINKED ISSUE #81: vecty.Render into non-body component

### LINKED ISSUE #247: RenderBody became blocking (it never returns) after PR #232
In [one of my applications](https://github.com/shurcooL/Go-Package-Store/blob/7f138b4c2e3ac1fd1ece09ab844716c76143293c/frontend/main.go#L43-L53), I've been using vecty's API roughly as follows:

```Go
vecty.RenderBody(body)
for {
    // sleep or wait for event ...
    vecty.Rerender(body)
}
```

It's possible I'm not using vecty API correctly, but that's what I came up with based on reading the documentation and experimenting (a while ago). It worked without noticeable issues until PR #232.

According to https://github.com/gopherjs/vecty/blob/master/doc/CHANGELOG.md#june-30-2019-pr-232-major-breaking-change, there have not been changes to `vecty.RenderBody` behavior.

The documentation of [`vecty.RenderBody`](https://godoc.org/github.com/gopherjs/vecty#RenderBody) says:

> RenderBody renders the given component as the document body. The given Component's Render method must return a "body" element.

I don't see anything that suggests that `vecty.RenderBody` would block execution and never return.

However, as of PR #232, that's what it does:

https://github.com/gopherjs/vecty/blob/2b6fc20f8913c7dadc6a67786be15c8c92bbd16a/dom.go#L1194-L1195

Which means my application never proceeds past the `vecty.RenderBody(body)` call.

Am I doing something wrong, or is this a bug/regression?

slimsag:
This PR adds support for rendering into arbitrary DOM nodes in the same way that rendering into the `body` worked previously, with one exception: blocking.

`RenderBody` remains blocking, since it effectively becomes a helper that 90% of Vecty applications will rely on, whereas `RenderInto` and `RenderIntoNode` which are introduced in this PR do not block. This is because:

a. Anyone rendering Vecty into arbitrary DOM nodes will likely want to do it more than once, so making the API of these functions non-blocking makes sense.
b. Anyone using these functions will be a more advanced user and understand when blocking is / is not needed.

In order to land this PR earlier for users who need this functionality, tests will not be added for the new functionality. They will be added in a follow-up as part of #168.

Fixes #81
Fixes #247


dom.go
```diff
@@ -1155,45 +1157,102 @@ func requestAnimationFrame(callback func(float64)) int {
...
 	batch.scheduled = true
-	nextRender, skip, pendingMounts := renderComponent(body, nil)
+	nextRender, skip, pendingMounts := renderComponent(c, nil)
 	if skip {
-		panic("vecty: RenderBody Component.SkipRender illegally returned true")
+		panic("vecty: " + methodName + ": Component.SkipRender illegally returned true")
 	}
-	if nextRender.tag != "body" {
-		panic("vecty: RenderBody expected Component.Render to return a body tag, found \"" + nextRender.tag + "\"")
+	expectTag := node.Get("nodeName").String()
```
dmitshur: I believe I'm running into an issue with `nodeName` case sensitivity here (see https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName and https://johnresig.com/blog/nodename-case-sensitivity/).

My program is doing:

```Go
// Initial frontend render.
err := vecty.RenderInto("body", body)
if err != nil {
	panic(fmt.Errorf("internal error: unexpected error from vecty.RenderInto: %v", err))
}
```

And getting:

```
Uncaught Error: internal error: unexpected error from vecty.RenderInto: vecty: RenderInto: expected Component.Render to return a "BODY", found "body"
```

I'm using HTML, Chrome 78.0.3904.108 on macOS.

dmitshur: I tried the following quick change to see if it would be sufficient to solve the problem:

```diff
diff --git a/dom.go b/dom.go
index 0eb000e..164a5a3 100644
--- a/dom.go
+++ b/dom.go
@@ -2,6 +2,7 @@ package vecty
 
 import (
 	"reflect"
+	"strings"
 )
 
 // batch renderer singleton
@@ -1225,7 +1226,7 @@ func renderIntoNode(methodName string, node jsObject, c Component) error {
 	if skip {
 		panic("vecty: " + methodName + ": Component.SkipRender illegally returned true")
 	}
-	expectTag := node.Get("nodeName").String()
+	expectTag := strings.ToLower(node.Get("nodeName").String())
 	if nextRender.tag != expectTag {
 		return ElementMismatchError{method: methodName, got: nextRender.tag, want: expectTag}
 	}
```

It was; after that, Go Package Store was rendering the frontend without issues.

I'm not sure if that's the optimal long term fix, or if there's a better way. You may not want to import a new `strings` package just for one function.

I tried using JavaScript's `toLowerCase` method, but for some reason that didn't work in the context of Vecty:

```
Uncaught Error: syscall/js: call of Value.Call on string
```

Even though it seems to work fine elsewhere, e.g., see https://gopherjs.github.io/playground/#/UxD9vDjbf0.

Hope this is helpful.

dmitshur: About the "syscall/js: call of Value.Call on string" error, I found it happens when using `syscall/js` package with GopherJS but not when using `github.com/gopherjs/gopherjs/js` package. So it's not Vecty-specific beyond Vecty using `syscall/js`.

It happens when using `syscall/js` with both GopherJS and Go/WebAssembly, so I filed golang/go#35917 to learn more about that.

slimsag: Thank you for investigating this! I've sent #251 to fix the issue on the Vecty side

PR #251, opened 2019-12-01
correctly handle the fact that nodeName is uppercase

slimsag:
Fixes issue reported by @dmitshur in https://github.com/gopherjs/vecty/pull/249/files#r352319035

### DIFF
--- a/dom_wasmjs_gopherjs.go
+++ b/dom_wasmjs_gopherjs.go
@@ -31,7 +31,13 @@ func RenderIntoNode(node js.Value, c Component) error {
 }
 
 func toLower(s string) string {
-	return valueOf(s).Call("toLowerCase").String()
+	// We must call the prototype method here to workaround a limitation of
+	// syscall/js in Go (but not GopherJS) where the following program will
+	// panic: https://gopherjs.github.io/playground/#/UxD9vDjbf0
+	//
+	// 	panic: syscall/js: call of Value.Call on string
+	//
+	return js.Global.Get("String").Get("prototype").Get("toLowerCase").Call("call", js.ValueOf(s))
 }
 
 var (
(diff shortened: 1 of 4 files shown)


slimsag:
@dmitshur thank you for reporting! I missed the fact that `nodeName` was uppercase. Would you be able to confirm if this fix works for you?

slimsag:
> panic: syscall/js: call of Value.Call on string

Ah, this actually appears to be an incorrect limitation of `GOOS=js GOARCH=wasm` in at least Go 1.12 and perhaps also in `master`

dom_wasmjs_gopherjs.go
```diff
@@ -30,6 +30,13 @@ func RenderIntoNode(node js.Value, c Component) error {
 	return renderIntoNode("RenderIntoNode", wrapObject(node), c)
 }
 
+func toLower(s string) string {
+	// We must call the prototype method here to workaround a limitation of
+	// syscall/js in both Go and GopherJS where we cannot call the
+	// `toLowerCase` string method. See https://github.com/golang/go/issues/35917
+	return js.Global().Get("String").Get("prototype").Get("toLowerCase").Call("call", js.ValueOf(s)).String()
```
dmitshur: Do you have a way of checking whether performance is impacted doing this on each `toLower` call, compared to:

```Go
func toLower(s string) string {
    return toLowerCase.Call("call", js.ValueOf(s)).String()
}

var toLowerCase = js.Global().Get("String").Get("prototype").Get("toLowerCase")
```

Also, the `string -> js.Value -> string` conversions can be short-circuited if you're willing to create a more dedicated helper for getting the node name, for example:

```Go
// +build js
func lowerNodeName(node jsObject) string {
    return toLowerCase.Call("call", node.Get("nodeName")).String()
}

// +build !js
func lowerNodeName(node jsObject) string {
    return strings.ToLower(node.Get("nodeName").String())
}
```

I don't have a sense whether the performance difference would be significant or negligible.

If performance isn't a bottneck or priority for now, this can all be considered later of course.

slimsag: I think even if performance of this is poor, it would probably be negligible AND since it only occurs on calls to `RenderBody`, `RenderInto`, `RenderIntoNode` it is not very frequently called, so as-is is probably fine for now.

Benchmarking things overall (all of Vecty) would definitely be a good idea, though that will probably come a bit later.

PR #274, opened 2020-12-27
Change markdown renderer to goldmark

nyiyui:
Goldmark is CommonMark compliant, and this is the same (or similar) reason it is being used in Hugo.

### DIFF
--- a/example/markdown/markdown.go
+++ b/example/markdown/markdown.go
@@ -5,7 +5,8 @@ import (
 	"github.com/hexops/vecty/elem"
 	"github.com/hexops/vecty/event"
 	"github.com/microcosm-cc/bluemonday"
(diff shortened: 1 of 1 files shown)


soypat:
Works A-OK. In my experience yuin's implementation works far better than blackfriday and it's also very well maintained. 
One slight downside is the binary size is greater (see below). I think it's worth the tradeoff. Worth mentioning the current import (github.com/slimsag/blackfriday) has a malformed go.mod file. See https://github.com/hexops/vecty/issues/305.

I'll update the readme next to reflect the size changes: wasm:3.6M->5.9M,  gz:1010k->1.37M,  brotli: 745k->952k

soypat:
Thank you @nyiyui for your contribution!

PR #294, opened 2022-01-16
update readme with links to examples and add supporting packages

soypat:
PTAL.

### DIFF
--- a/doc/projects-using-vecty.md
+++ b/doc/projects-using-vecty.md
@@ -14,7 +14,7 @@ This is just a list of applications using Vecty that we're aware of. We don't st
 
 Know of a project using Vecty that is not on this list? Please [let us know](https://github.com/hexops/vecty/issues/new)!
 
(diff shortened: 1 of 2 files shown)


doc/projects-using-vecty.md
```diff
@@ -13,3 +13,11 @@ This is just a list of applications using Vecty that we're aware of. We don't st
 | [Vocdoni Block Explorer](https://gitlab.com/vocdoni/vocexplorer) | A custom blockchain explorer for decentralized digital governance toolkit [Vocdoni.io](https://Vocdoni.io). |
 
 Know of a project using Vecty that is not on this list? Please [let us know](https://github.com/hexops/vecty/issues/new)!
+
+The following are user created packages for aiding with frontend development using vecty.
```
slimsag, line 17:
```suggestion
## Libraries
```

How about a header instead?

